PR #222, opened 2019-10-25
Add base64 tab to connection strings

wbobeirne:
### Description

In addition to hex encoding, some projects use base64 encoded versions of the certs and macaroons (cough cough like my own node lapp guides.) This just adds that as one of the options. Refactors the `HexStrings` component to take encoding type as a string, and generecizes some of the tex. Spanish may need some love.

Let me know if you would like to see tests here, didn't add any!

### Steps to Test

1. Flip between hex and base64, confirm they both decode to the same values.

### Screenshots

Screen Shot 2019-10-25 at 5 20 24 PM


jamaljsr:
Wow!! a PR already.. appreciate it sir.

This is a good addition. I was unaware that some apps used base64, but the more compatibility with others, the better. I'll review the code over the weekend and give feedback.

Regarding the tests, I want to keep code coverage as high as reasonably possible. Been keeping it at 100% so far and don't want to break the trend so soon. No worries though if you don't want to add them. I can do them. Maybe sometime in the future, I'll require it from contributors but you get a pass for being first. 💪 I have to figure out why the CI build didn't run and scream at you for dropping the coverage. It looks like you found another bug. Thanks for that too

jamaljsr:
Ignore my last comment about pushing to your branch. I just merged this and will make my changes in a separate PR. Thanks gain

PR #245, opened 2019-11-14
fix(i18n): correct spanish translations

federicobond:
Congrats on the app, it's works pretty great so far!


bolatovumar:
@jamaljsr do you use CrowdIn like Zap does (https://crowdin.com/project/zap-desktop)? Should make these sort of contributions easier.

jamaljsr:
@federicobond thank you so much for fixing these strings. I don't know any spanish and relied on Google Translate for the translations. I just added spanish because I wanted to have a second language to ensure the UI updated properly. 

@bolatovumar yes, I intend to integrate crowdin into the workflow with the next release. I structured the locale files so that they are already compatible with crowdin, so it shouldn't be too difficult to setup.

PR #247, opened 2019-11-20
feat(ui): disable "Close channel" button if opening channel to self

### LINKED ISSUE #246: Disable "Open channel" button when same nodes selected
Currently you can attempt to open a channel to self. You get an error notification in this case but this should probably not be allowed in the first place.

bolatovumar:
Closes #246


bolatovumar:
@jamaljsr no problem.

PR #251, opened 2019-11-27
feat(network): add support for c-lightning v0.7.3

jamaljsr:
### Description

This PR adds support for c-lightning v0.7.3 nodes. All of the existing functionality for LND nodes has been implemented for c-lightning as well (ex: open channels, launch terminal, etc). 

**Note:** c-lightning syncs chain data with the bitcoind node once every 30 secs. So most actions are not reflected immediately in the UI. Polar needs to wait 30s then query the node for updated channels & balances.

There is still a bit more work I have to finish before merging. This early PR is for anyone interested in c-lightning support to test and provide feedback.   

Remaining Improvements:
- [x] fix some c-lightning channels not being displayed
- [x] try to improve the UX around the c-lightning 30s chain sync delay
- [x] add and update unit tests
- [x] test full functionality on Windows & Linux

### Steps to Test

1. Create a network with 2x c-lightning and 1x LND nodes
2. Deposit funds and confirm balance updates correctly
3. Launch the Terminal and run a few lightning-cli commands
4. Open a channel to a c-lightning and a LND node
5. Close one of the channels
6. Remove the node from the network
7. Drag a new c-lightning node from the sidebar
8. Repeat steps 2-5

### Screenshots

image


jamaljsr:
Leaving this note here for my future self.

I have spent a ton of time trying to get the c-lightning docker container to mount a shared volume on Windows with no success thus far. This is a common problem with some linux-based applications, such as postgresql ([good explanation](https://forums.docker.com/t/data-directory-var-lib-postgresql-data-pgdata-has-wrong-ownership/17963/24)), being unable to write to the docker volume mounted on an NTFS file system. It can work when run as root, but not when run as a normal user. Unfortunately, Linux needs the container to run rootless due to file permission issues. So I may need to add some conditional logic in the docker files based on the host OS.

I will revisit c-lightinng on Windows in the future, as its already been a few weeks working on this branch and I'd like to get some on other features done for the next release. For now, I have simply disabled the ability to add c-lightning nodes on Windows. I have tested and confirmed everything works well on Mac and Linux.

Here's a [gist](https://gist.github.com/jamaljsr/404c20f99be2f77fff2d834e2449158b) with the error logs and docker files that may be helpful later.

PR #278, opened 2020-01-18
fix(ui): make SidebarCard scrollable

### LINKED ISSUE #277: Sidebar card is not scrollable
**Describe the bug**
I'm using Polar on Mac Air 13″ not in full screen so the sidebar card is not scrollable at all. As a result I can't see all the details about my nodes unless I go into full-screen mode or something. Would be nice not to have to do that.

**To Reproduce**
1. Open app in a small window
2. Create and start a network
3. Try to scroll to see more details in the sidebar card

**Expected behavior**
Able to scroll through the side card and see all details for a node.

**Screenshots**
Screen Shot 2020-01-17 at 8 41 03 PM

 - OS: macOS Mojave
 - Polar version: 0.2.0

bolatovumar:
Closes #277

### DIFF
--- a/src/components/designer/SidebarCard.tsx
+++ b/src/components/designer/SidebarCard.tsx
@@ -9,5 +9,25 @@ export default styled(Card)`
   width: 360px;
   border-radius: 2px;
   box-shadow: 0 10px 20px rgba(0, 0, 0, 0.1);
-  overflow: hidden;
+  overflow: auto;
+
+  &::-webkit-scrollbar {
+    width: 8px;
+    background-color: rgba(0, 0, 0, 0);
+    border-radius: 10px;
+  }
+
+  &::-webkit-scrollbar:hover {
+    background-color: rgba(0, 0, 0, 0.1);
+  }
+
+  &::-webkit-scrollbar-thumb:vertical {
+    background-color: rgba(0, 0, 0, 0.2);
+    border-radius: 10px;
+  }
+
+  &::-webkit-scrollbar-thumb:vertical:active {
+    background-color: rgba(0, 0, 0, 0.6);
+    border-radius: 10px;
+  }
 `;


jamaljsr:
I was aware of this issue but intentionally didn't fix it using `overflow: auto`. For one, I assumed not many developers run on screen sizes that small. Secondly, the default scrollbar UI varies per OS. While this fix works great on Mac, it looks very ugly on Linux & Windows. 

image

image

I would prefer to utilize a library to make the scrollbars consistent across OS's and match the UI of the app (light/dark). I've used [react-custom-scrollbars](https://malte-wessel.com/react-custom-scrollbars/) in other projects and have been happy with it. I just haven't gotten around to adding it to Polar.  

If you would like to update this PR to use the custom scrollbars, it would be appreciated. If not, then I will still merge this PR but keep in mind I plan to improve it further before the next major release. Let me know how you'd like to proceed.

Thanks so much for the contribution.

wbobeirne:
You could also just use the webkit scrollbar css selectors to style it since you don't need to worry about cross browser support given that it's electron: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

jamaljsr:
Great suggestion @wbobeirne! I keep forgetting that we only use Chromium. Hard to break the webdev mentality. Styling the scrollbars is certainly a better approach than adding another dependency. Thank you very much sir 👌

bolatovumar:
@jamaljsr i can re-style the scrollbar and see if that works. If it doesn't work then I can go the custom scrollbar route.

jamaljsr:
@bolatovumar that sounds great! thank you

bolatovumar:
Added Mac-style scrollbar styling which works well both in dark/light theme and Mac/Windows.

PR #279, opened 2020-01-18
docs: fix spelling and add Polar version field to bug template

bolatovumar:
Minor spelling error fix and added Polar version field to bug template document.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,9 +24,10 @@ A clear and concise description of what you expected to happen.
 If applicable, add screenshots to help explain your problem.
 
 **Desktop (please complete the following information):**
- - OS: [e.g. OSX Majave or Windows 10 v1903]
- - Docker Version: [e.g. 19.03.1]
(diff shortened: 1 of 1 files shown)


bolatovumar:
@jamaljsr looks like the pre-commit hook formats markdown files and botches them. I had to turn off pre-commit hooks to get the change committed without messing up the current formatting of the file.

jamaljsr:
> @jamaljsr looks like the pre-commit hook formats markdown files and botches them. I had to turn off pre-commit hooks to get the change committed without messing up the current formatting of the file.

I'll investigate the linter formatting for markdown files. Thanks for letting me know.

jamaljsr:
@bolatovumar just following up. I looked into the markdown formatting issue and found that it's being caused by prettier. These markdown files contain metadata (called front matter) at the top of the file. Prettier is supposed to ignore this content based on their docs & issues, but It doesn't appear to be working as advertised. 

To fix the problem for now, I have added these files to the .prettierignore file in 5267f248fdee77a6d0a526b42b6bcb762733d932, so that they won't be messed up by the pre-commit hook.

PR #285, opened 2020-01-25
fix: handle errors in Bitcoind logger

### LINKED ISSUE #284: Console error when starting a network
See the screenshot below for the error I see in the console when starting a new network:

Screen Shot 2020-01-23 at 8 10 11 PM

As you can see from the log the `data` parameter in `log()` method doesn't have a `body` in request (see the log of the request in the top section of the screenshot).

Not sure if this error matters at all but thought I would bring it up.

bolatovumar:
Closes #284

### DIFF
--- a/src/lib/bitcoin/bitcoindService.spec.ts
+++ b/src/lib/bitcoin/bitcoindService.spec.ts
@@ -75,6 +75,34 @@ describe('BitcoindService', () => {
     expect(spy).toBeCalledWith('BitcoindService: [response]', '{}');
   });
 
+  it('should log correctly for rpc response error', async () => {
+    const spy = jest.spyOn(log, 'debug');
+    mockBitcoin.mockImplementationOnce(options => {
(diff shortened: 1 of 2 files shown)


jamaljsr:
This looks good from just glancing at the code. I'll test it locally in the next day or two.

I think this update will reduce test coverage due to the new if branches not being exercised. If you want to try to write some tests, you can use `yarn test:ci` to view the coverage. If you don't want to be bothered, no worries. I will still merge this and write the tests afterwards.

I have to fix the test coverage not being reported in PR's like this from forked repos.

bolatovumar:
@jamaljsr i tested it locally and this fixes the error. I can write some tests for it.

bolatovumar:
@jamaljsr alright, added a couple of tests

jamaljsr:
Awesome @bolatovumar! Thanks a ton.

PR #289, opened 2020-02-04
fix: make sure localBalance is never undefined for pending channel

### LINKED ISSUE #287: Source balance shown as "undefined sats" before channel is confirmed
**Describe the bug**
`Source Balance` shows `undefined sats` before channel is confirmed if channel was opened outside of Polar UI.

**To Reproduce**
1. Run RTL locally
2. Open a channel from the LND node you using for RTL to another LND node in Polar
3. Notice that `Source Balance` for the channel is shown as `undefined sats`

**Expected behavior**
It should probably show 0 for `Source Balance`.

**Screenshots**
Screen Shot 2020-02-02 at 7 41 43 PM

- OS: macOS Mojave
- Polar Version: v0.2.1

**Additional context**
The problem goes away once you mine enough blocks to have the channel confirm.

bolatovumar:
fix #287

### DIFF
--- a/src/shared/lndDefaults.ts
+++ b/src/shared/lndDefaults.ts
@@ -149,26 +149,35 @@ export const defaultLndWaitingCloseChannel = (
 
 export const defaultLndPendingChannels = (
   value: Partial<LND.PendingChannelsResponse>,
-): LND.PendingChannelsResponse => ({
-  totalLimboBalance: '',
-  pendingOpenChannels: mapArray(
-    value.pendingOpenChannels || [],
-    defaultLndPendingOpenChannel,
-  ),
-  pendingClosingChannels: mapArray(
-    value.pendingClosingChannels || [],
-    defaultLndClosedChannel,
-  ),
-  pendingForceClosingChannels: mapArray(
-    value.pendingForceClosingChannels || [],
-    defaultLndForceClosedChannel,
-  ),
-  waitingCloseChannels: mapArray(
-    value.waitingCloseChannels || [],
-    defaultLndWaitingCloseChannel,
-  ),
-  ...value,
-});
+): LND.PendingChannelsResponse => {
+  const {
+    pendingOpenChannels,
+    pendingClosingChannels,
+    pendingForceClosingChannels,
+    waitingCloseChannels,
+    ...rest
+  } = value;
+  return {
+    totalLimboBalance: '',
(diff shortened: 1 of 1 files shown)


jamaljsr:
Thank for finding this bug and proposing a fix. I would prefer to fix this at the root of the problem instead of putting a bandaid on it. 

I traced the source back to the [defaultLndPendingChannels](https://github.com/jamaljsr/polar/blob/master/src/shared/lndDefaults.ts#L150) function. The purpose of the functions in this file is to provide default values for all LND responses since GRPC omits falsey values. I can see that the issue in this function is that the `...value` spread is overwriting the new `pendingOpenChannels` array with the old one from `value`. So the default value `0` for `remoteBalance` (set in `defaultLndPendingOpenChannel`) is not being returned.

To fix this, we need to be sure that the new arrays are not overwritten. I suggest using the following code:
```js
export const defaultLndPendingChannels = (
  value: Partial<LND.PendingChannelsResponse>,
): LND.PendingChannelsResponse => {
  const {
    pendingOpenChannels,
    pendingClosingChannels,
    pendingForceClosingChannels,
    waitingCloseChannels,
    ...rest
  } = value;
  return {
    totalLimboBalance: '',
    pendingOpenChannels: mapArray(
      pendingOpenChannels || [],
      defaultLndPendingOpenChannel,
    ),
    pendingClosingChannels: mapArray(
      pendingClosingChannels || [],
      defaultLndClosedChannel,
    ),
    pendingForceClosingChannels: mapArray(
      pendingForceClosingChannels || [],
      defaultLndForceClosedChannel,
    ),
    waitingCloseChannels: mapArray(
      waitingCloseChannels || [],
      defaultLndWaitingCloseChannel,
    ),
    ...rest,
  };
};
```

If you can update this PR to use the code above, I will merge it.

bolatovumar:
@jamaljsr oh, I see, that makes sense. Interestingly enough that function never gets hit in my debugger and I just assumed it wasn't the issue.

jamaljsr:
This function wouldn't get hit by your debugger because it's being executed by the main electron process, not the renderer process. I'm guessing the debugger is only attaching to the chromium renderer, where mostly all of the application logic is run. I tried to keep the main process as slim and dumb as possible.

bolatovumar:
@jamaljsr oh, ok, makes sense

PR #295, opened 2020-02-11
Add import/export functionality

torkelrogstad:
In this PR we add import/export functionality. 

Working: 
 - Add dropdown to network (where "rename" and "delete" are) for exporting. This zips a JSON file with the current network and prompts the user for where to store the zip
- Add new "import network" page where user can select a zip file, and import this

Todo: 
 - [ ] Remove user name from stringified network in zip?
 - [ ] Tests
 - [x] Docs
 - [x] indicator while we're importing/exporting networks
 - [x] Adding newly imported networks to chart and persisting this to disk
 - [x] modal saying we successfully imported network


jamaljsr:
@torkelrogstad I won't do a full review because you marked it as a draft, but I have skimmed through the code and ran it. Overall, it works as advertised, which is awesome. Thanks for working on this feature and let me know when you feel ready for me to do a thorough review.

torkelrogstad:
This PR is now in a reviewable state, @jamaljsr. I've added some tests where I mimicked existing tests. The hairy parts of this PR (the zipping/unzipping/`networks.json` tweaking) is not tested, though... LMK what you think is important to test here. I'd also appreciate some pointers on how to approach that, w.r.t. best practices around tests that require real file system IO. 

I also didn't tackle the first item on my checkbox from above, removing the user name of the user that exported the network. This would take some effort, and doesn't seem to be terribly important. 

Here's an exported network you can pull down and import: [polar-foo.zip](https://github.com/jamaljsr/polar/files/4214525/polar-foo.zip)

jamaljsr:
Hey @torkelrogstad, I will give the code a thorough review by tomorrow. 

Regarding testing, the simplest way to determine what tests to add is to look at the code coverage report. If you run `yarn test:ci` it will output a list of all of the source files along with the line numbers which are not covered by tests. After that, you can also open the `/coverage/lcov-report/index.html` file in a browser to see a more visual view of what code has no coverage. This tells you exactly where you should add more tests to cover all possible code paths. I do have coverage reports posting to PRs (ex: [282](https://github.com/jamaljsr/polar/pull/282#issuecomment-576832266)) but unfortunately this feature doesn't work yet with GitHub Actions on PRs from forked repos.

Using coverage as a guide also helps you organize your code better. For example, I see that you have a fair amount of logic written in the NetworkView component and button click handlers. It would be challenging to write tests for these since the testing framework has to interact with the UI. It is a better practice to keep your components as small as possible and have util functions for the heavy lifting. The only logic in these files should be to determine what to render on the screen, with little to none of the plumbing necessary to perform an action. This also allows you to reuse that functionality in other components, which keeps your codebase [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Since this code would be difficult to test, you would be encouraged to move it to a separate file. Then you'd have a lot more flexibility to pass in arguments to the function to induce errors and handle edge cases.

I know testing feels like a pain upfront, but I can tell you from experience that it leads to a much more developer friendly codebase in the future after years of changes.

jamaljsr:
I have started reviewing this PR by trying to break it in as many ways as I can think of. Here are some issues I have found by 

1. The `polar-foo.zip` file that you posted in your comment appears to be corrupt. I receive an error when I try to import it as well as when I try to extract it in Finder.
   ```
   Could not unzip /Users/jamal/Downloads/polar-foo.zip into /var/folders/zq/b69vg6b50x3gxlqfxtm6vybw0000gn/T/polar-foo: Error: invalid distance too far back
    at Zlib.zlibOnError [as onerror] (zlib.js:170)
   ```
   image
1. I created a network but did not start it. I then exported it, deleted it, then tried to import it. I received this error in the console.
   ```
   Uncaught (in promise) Error: ENOENT: no such file or directory, stat '/var/folders/zq/b69vg6b50x3gxlqfxtm6vybw0000gn/T/polar-import/volumes'
   ```
1. I created a network, started it, then added some channels and made payments. While the network is running, I exported it. I then stopped it, deleted it, and imported it back in. Polar believes that the network is still running but it cannot connect to the nodes. Obviously the docker containers are not running for a newly imported network. I then stopped the network, and started it up again. The bitcoin node came online but the LND & c-lightning nodes did not. My high-level assumption is that the node's data is now corrupt because Polar zipped up the folders while the nodes were running. They didn't gracefully shutdown properly. I suggest a simple fix of not allowing the export of running networks. Just show an error notification if the user clicks Export while the network is running.

1. I exported a network with a c-lightning node from a Mac and transferred the zip file to my Windows machine. I was able to import the network and start it but the c-lightning node failed to start. This is because our c-lightning docker containers are not supported on Windows yet. We should prevent importing networks with c-lightning nodes on Windows.

I also have some cosmetic suggestions to improve the UI
1. On the Get Started screen, I don't think it is necessary to have the Import button in the center. This screen is only shown to first-time users. I think it is highly unlikely that importing a network would be a common use case for folks just starting, so it doesn't need to be this prominent in the UI. There is also the Import link in the top nav that is easily visible for those that need it.
1. We should add an icon to the "Import Network" link in the top navbar. It looks a bit strange without one. I would suggest `<ImportOutlined />` from antd.
1. On the Import screen:
   - I think it would look better if the Card with the dropzone filled the full height of the screen. This should be doable with a bit of flexbox CSS.
   - since the user will only ever upload one file, I don't think it's necessary to have them select a file then click on the Import button. The import process should immediately initiate after the file is dropped or selected. This would remove the need to show the Remove and Import buttons.
   - there should be some sort of loader displayed while the network it being imported. It can take a few seconds and when I did it, I wasn't sure if anything was happening until the success modal popped up.

This is all the feedback I have from a user's perspective. I will go ahead and dive into the code now and provide feedback on that separately.

src/utils/network.ts
```diff
@@ -173,6 +177,93 @@ export const createBitcoindNetworkNode = (
...
+
+  return chart;
+};
+
+/**
+ * @returns The newly created network, the chart corresponding to it
+ *          and the destination of the unzipped files
+ */
+export const getNetworkFromZip = async (
+  zip: string,
```
jamaljsr: nit: rename to `zipPath` so that it is clear that this is a file path

src/utils/network.ts
```diff
@@ -173,6 +177,93 @@ export const createBitcoindNetworkNode = (
...
+  newId: number,
+): Promise<[Network, IChart, string]> => {
+  const destination = join(os.tmpdir(), basename(zip, '.zip'));
+  await files.unzip(zip, destination).catch(err => {
+    console.error(`Could not unzip ${zip} into ${destination}:`, err);
```
jamaljsr: Use `electron-log` to log errors instead of `console` so that it gets saved to the log file on disk when using the production build.

src/utils/network.ts
```diff
@@ -173,6 +177,93 @@ export const createBitcoindNetworkNode = (
...
+    throw err;
+  });
+
+  const [network, chart] = await Promise.all([
```
jamaljsr: I prefer to use `await`'s over `Promise.all()` when possible, which makes the code easier to read.

src/utils/files.ts
```diff
@@ -33,6 +41,147 @@ export const exists = async (filePath: string): Promise<boolean> =>
...
+    // we've added all files, tell this to the archive so it can emit the 'close' event
+    // once all streams have finished
+    archive.finalize();
+  });
+};
+
+/**
+ * Unzips `zip` into `destination`
+ */
+export const unzip = (zip: string, destination: string): Promise<void> => {
```
jamaljsr: nit: rename `zip` to `zipPath` to be more specific

src/utils/files.ts
```diff
@@ -33,6 +41,147 @@ export const exists = async (filePath: string): Promise<boolean> =>
...
+ * volumes            // directory with all data files needed by nodes
+ * network.json       // serialized network object
+ * chart.json         // serialized chart object
+ * ```
+ *
+ * @return Path of created `.zip` file
+ */
+export const zipNetwork = async (network: Network, chart: IChart): Promise<string> => {
+  return new Promise(async (resolve, reject) => {
+    const zipPath = join(os.tmpdir(), `polar-${network.name}.zip`);
```
jamaljsr: Should sanitize the network name since it could contain special characters that aren't allowed in file names (ex: /, ?, <, >, \, :, *, etc.)

src/components/network/NetworkView.tsx
```diff
@@ -150,6 +190,17 @@ const NetworkView: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
...
             }}
             onDeleteClick={() => showRemoveModal(network.id, network.name)}
+            onExportClick={async () => {
+              const destination = await exportNetwork(network);
+              if (!destination) {
+                return;
+              }
+
+              Modal.success({
+                title: 'Exported network',
```
jamaljsr: These strings should be saved in the `en-US.json` file

src/components/network/ImportNetwork.tsx
```diff
@@ -0,0 +1,137 @@
...
+                ),
+              );
+
+              networkActions.add(newNetwork);
+              designerActions.setChart({ chart, id: newId });
+              await networkActions.save();
+
+              log.info('imported', newNetwork);
+              Modal.success({
+                title: 'Imported network',
```
jamaljsr: These strings should be saved in the `en-US.json` file

src/components/network/ImportNetwork.tsx
```diff
@@ -0,0 +1,137 @@
...
+          <p className="ant-upload-text">{l('fileDraggerArea')}</p>
+        </Upload.Dragger>
+        <Styled.ButtonContainer>
+          <Button onClick={() => setFile(undefined)} disabled={file === undefined}>
+            {l('removeButton')}
+          </Button>
+          <Button
+            onClick={async () => {
+              if (!file) {
+                throw Error('File was undefined in import submit function!');
```
jamaljsr: If this error is thrown it will crash the app, it would be to display an error notification instead.

torkelrogstad: This is more of an invariant that should always hold, though. If the file is undefined, the button should be disabled. So if this point is ever reached, we have a bug. I think it's good practice to be explicit about errors that should never happen (as opposed to errors that could very well happen, these should be handled gracefully). What's your thoughts on this?

src/utils/network.ts
torkelrogstad: The benefit of `Promise.all` is that it lets you await on promises that are run concurrently. If we did

```typescript
const network = await readNetwork(...)
const chart = await readChart(...)
```
the chart-reading function wouldn't start until the network-reading function finished. This would be equivalent performance-wise:

```typescript
const networkPromise = readNetwork(...)
const chartPromise = readChart(...)
const network = await networkPromise
const chart = await chartPromise
```

I'd say using `Promise.all` is more readable than :arrow_up:. Feel free to disagree though, in that case I'll change it.

jamaljsr: You have a valid point regarding the concurrency of `Promise.all` so I'll concede that it is fine to use it in this case.

src/components/network/ImportNetwork.tsx
jamaljsr: I would argue that throwing errors for situations that should never happen is a bit misleading. The fact that you are explicitly checking for this case makes me believe that it is possible when reading the code. Using this conditional also means you should have a unit test to exercise this logic, which would be difficult to do since the button is disabled. I think it's better to just cast the object `file as RcFile` to make Typescript happy. As someone reading the code, the cast tells me that `file` should never be undefined, which is true.

e2e/Import.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+  .page(pageUrl)
+  .beforeEach(Home.clickImportButton)
+  .afterEach(assertNoConsoleErrors)
+  .afterEach(cleanup);
+
+test('should be on the import network route', async t => {
+  await t.expect(getPageUrl()).match(/network_import/);
+});
+
+test('when the user aborts the file dialog, nothing should happen', async t => {
```
torkelrogstad: I'm having issues with this test. I can't get Testcafe to trigger the file dialog... Any idea what I'm doing wrong?

src/store/models/network.spec.ts
```diff
@@ -17,6 +18,28 @@ jest.mock('utils/files', () => ({
...
+    return jest.fn().mockImplementation(() => {
+      const network = {
+        id: 1,
+        nodes: {
+          bitcoin: [{}],
+          lightning: [{}],
+        },
+      };
+      return [network, {}];
+    })();
```
torkelrogstad: This syntax feels a bit weird (having to do `()` at the end), LMK if there's anything really strange I'm doing with Jest here

src/utils/zip.ts
```diff
@@ -0,0 +1,109 @@
...
+};
+
+export const zip = ({ destination, objects, paths }: ZipArgs): Promise<void> =>
+  new Promise(async (resolve, reject) => {
+    const output = fs.createWriteStream(destination);
+    const archive = archiver('zip');
+
+    // finished
+    archive.on('finish', () => resolve());
+    archive.on('error', err => {
```
torkelrogstad: I'm having some issues with testing the `error` and `warning` branches here. How would I solve this with a mock?

src/components/network/ImportNetwork.tsx
```diff
@@ -0,0 +1,95 @@
...
+      .catch(error => {
+        notify({ message: l('importError', { file: file.name }), error });
+      })
+      .then(() => {
+        setImporting(false);
+      });
+
+    // return false to prevent the Upload.Dragger from sending the file somewhere
+    return false;
+  };
```
torkelrogstad: This function is not tested in unit tests. One solution here could be to create a hidden element on the page that has a `onClick` set to this, and then assert on that in tests. Any thoughts on that?

src/utils/zip.ts
jamaljsr: This was a bit tricky because the `archiver` package exposes its main function as the default export. After a little tinkering, I've gotten a couple tests passing which simulate the `error` and `warning` events. 
1. The `jest.mock('archiver`, ...);` code should be placed in a global mock file because it should be mocked in all tests. I just put it all in one file for easier pasting.
1. I mocked `fs` in these tests because your code used that module. I suggest switching to `fs-extra` which already has a global mock in __mocks__. It would just need to be updated with any missing functions declared there, like `createWriteStream`.
```typescript
import fs from 'fs';
import * as archiver from 'archiver';
import { PassThrough } from 'stream';
import { zip } from './zip';

jest.mock('fs');
jest.mock('archiver', () => {
  // TODO: this code should live in __mocks__/archiver.js

  // eslint-disable-next-line @typescript-eslint/no-var-requires
  const { PassThrough } = require('stream');
  let mockStream: any;
  // return a fake stream when "archiver()" is called in the app
  const ctor = function() {
    mockStream = new PassThrough();
    mockStream.file = jest.fn();
    mockStream.directory = jest.fn();
    mockStream.append = jest.fn();
    mockStream.finalize = jest.fn();
    return mockStream;
  };
  // attach a func to emit events on the stream from the tests
  ctor.mockEmit = (event: string, data: any) => mockStream.emit(event, data);

  return ctor;
});

const fsMock = fs as jest.Mocked<typeof fs>;
const archiverMock = archiver as jest.Mocked<any>;

it('should fail if there is an archiver error', async () => {
  fsMock.createWriteStream.mockReturnValueOnce(new PassThrough() as any);

  const promise = zip({
    destination: 'path-to-zip',
    objects: [],
    paths: [],
  });

  // emit an error after a small delay
  const mockError = new Error('test-error');
  setTimeout(() => {
    archiverMock.mockEmit('error', mockError);
  }, 100);

  await expect(promise).rejects.toEqual(mockError);
});

it('should fail if there is an archiver warning', async () => {
  fsMock.createWriteStream.mockReturnValueOnce(new PassThrough() as any);

  const promise = zip({
    destination: 'path-to-zip',
    objects: [],
    paths: [],
  });

  // emit an error after a small delay
  const mockError = new Error('test-warning');
  setTimeout(() => {
    archiverMock.mockEmit('warning', mockError);
  }, 100);

  await expect(promise).rejects.toEqual(mockError);
});
```

tsconfig.json
```diff
@@ -16,5 +16,6 @@
     "jsx": "preserve",
     "baseUrl": "src"
   },
-  "include": ["src"]
+  "include": ["src"],
+  "exclude": ["src/utils/tests/resources"]
```
jamaljsr: These resource files should be removed since the tests shouldn't be using the file system

.eslintignore
```diff
@@ -7,3 +7,6 @@
 
 # compiled by tsc from /src/electron/
 /public
+
+# rest resources
+/src/utils/resources
```
jamaljsr: I don't think this is necessary

jamaljsr:
Hey @torkelrogstad are you still working on this PR? I want to package up a new release this coming week and would love to include this feature. If you don't have time to finish this up now, I am happy to work on it myself. I just don't want to take over if you plan to continue with it yourself.

torkelrogstad:
@jamaljsr Sorry for being a bit off the grid for the last days. Hoping to finish it up today, working on addressing your latest round of review now.

jamaljsr:
Ok great. Thanks for letting me know.

torkelrogstad:
I pushed what I had time to do today. I've addressed some of your comments, but did not get around to all of them. If you want to get this feature wrapped up so you can do a release, feel free to push changes here.

jamaljsr:
Ok, this looks pretty close to being ready. I'll finish it up in a day or two. Thanks for contributing what you have so far.

jamaljsr:
@torkelrogstad I finished up this PR and will merge it. Thank so much for your work on this feature. 👍

PR #381, opened 2020-09-08
fix(lnd): fixes clipboard copy value for p2pLnUrlInternal

Jasonvdb:
### Description
Copying the `P2P Internal` value to the clipboard was copying the value for `P2P External` instead.

### Steps to Test

1. Setup a network
2. Copy the value for `P2P Internal`
3. Paste the value somewhere (should match the displayed value in the app)


jamaljsr:
Good catch. And thanks for the fix. 🥇 

Can you just change the commit message to follow the [commitlint](https://github.com/conventional-changelog/commitlint#what-is-commitlint) standard.

Example
```
fix(lnd): fixes clipboard copy value for p2pLnUrlInternal
```

Jasonvdb:
Done 👌

jamaljsr:
Perfect! Thanks you

PR #390, opened 2020-10-26
custom nodes: update to work with latest eclair

halseth:
Updates instructions for how to build custom eclair image to work with latest version.

The other Dockerfile probably also needs update. And the config option
```
 --on-chain-fees.max-feerate-mismatch=100000
```

should now be
```
  --on-chain-fees.feerate-tolerance.ratio-low=0.00001
  --on-chain-fees.feerate-tolerance.ratio-high=10000.0
```


jamaljsr:
Wow! I was expecting the diff to be much larger. I'll test it out myself and merge if it's all good. Thanks so much for the PR.

jamaljsr:
tACK 🚀

squashing to fix commit message

PR #414, opened 2020-12-29
LND app example

### LINKED ISSUE #408: Feature Request: An example of using polar
Firstly, I just want to say that Polar is awesome and I'm really excited that a tool like this exists for the ecosystem!

I've been playing with it for the last few days, trying to get a working script that allows Alice to open a channel with Bob. Doing so, at least in my chosen language (TypeScript) was trickier than I thought it'd be.

For new developers who are unsure where to get started, a simple working example in the repo would go a long way. What I'd like to propose is this script to an `examples` folder in the repo, which would looks something like this:

```ts
import ln from 'ln-service';
import { ALICE_LND_AUTH } from './config';

const openChannel = async (lnd: ln.AuthenticatedLND) => {
  const { peers } = await ln.getPeers({ lnd });
  const args: ln.OpenChannelArgs = {
    local_tokens: 100000,
    partner_public_key: peers[0].public_key,
  };
  return ln.openChannel({ lnd, ...args });
};

(async () => {
  const { lnd } = ln.authenticatedLndGrpc(ALICE_LND_AUTH);
  try {
    const { pending_channels } = await ln.getPendingChannels({ lnd });
    if (!pending_channels.length) {
      const { transaction_id } = await openChannel(lnd);
      console.log('No pending channels found, opened one: ', transaction_id);
    }
    console.log('Pending channels found: ', pending_channels.length);
  } catch (err) {
    console.error(err);
  }
})();
```

It's a really simple script, so it might not seem all that useful, but I think it'd be a good starting point for a couple of reasons:
- Having an example of a working TypeScript gRPC client is good for new developers. The state of various LND gRPC libraries' support for TypeScript requires some hacking to get type definitions (in my case, using [`ln-service`](https://github.com/alexbosworth/ln-service) + [`patch-package`](https://www.npmjs.com/package/patch-package) to add types from a [not-yet-merged PR](https://github.com/alexbosworth/ln-service/pull/135)). So this project would be a useful way to illustrate the bootstrapping needed for a new developer who wants to use a popular language like TS.
- Showing how the outputs from Polar (e.g. gRPC connection info, TLS, macaroons) are meant to be used to authenticate with a lightning client

In the future, such an example would ideally be extended to a simple interactive interface that lets developers play with introducing real app-like functionality.

I'm happy to clean up what I have and put it in PR, but want to get some feedback before I do so and see if this would be useful.

murtyjones:
Closes #408

### Description

This PR adds an implementation of an example TypeScript app for `lnd` nodes that a developer might be working on.

### Steps to Test

Contained within `README.md`

### Screenshots

(The app commands have been changed a bit since this video, but it gives the general idea):

https://user-images.githubusercontent.com/708562/103317432-21578400-49f9-11eb-81d7-3f015e41a416.mov


examples/typescript/lnd/tsconfig.json
```diff
@@ -0,0 +1,70 @@
+{
```
murtyjones, line 1: this is largely the defaults from `tsc init`.

I'll clean it up and remove the comments in the 90% version of the PR.

examples/typescript/lnd/src/nodeHandler.ts
```diff
@@ -0,0 +1,45 @@
+import { LnRpc, Peer, WalletUnlockerRpc } from '@radar/lnrpc';
+
+export default class NodeHandler {
+  constructor(private lnrpc: LnRpc & WalletUnlockerRpc) {}
```
murtyjones, line 4: 1. Dependency injection is nice for testing but I wonder whether, in such a small example, it really makes sense to use it here like the PR currently does.
2. With some abstraction, this app could work with multiple Lightning client implementations (`lnd`, `eclair`, etc.) but I'm torn on whether that's a good idea, or if it's better to have three separate but similar apps that use the different clients

examples/typescript/lnd/src/nodeHandler.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  seePeerCount = async () => {
+    const info = await this.lnrpc.getInfo();
+    return info.numPeers;
+  };
+
+  handlers = {
+    'See number of peers': this.seePeerCount,
+    'List peers': this.listPeers,
```
murtyjones, line 20: Open to having better actions here if that matters, I just picked these because they were easy to call. but maybe more intricate actions would be better

examples/typescript/lnd/src/nodeHandler.ts
jamaljsr: 1. I think it's good to pass in the dependency
2. This would add a bunch of complexity for a basic example, so I think it's best to have separate apps for each LN implementation

examples/typescript/lnd/src/nodeHandler.ts
jamaljsr: It would be good to list and open channels. These are a common use case for most apps building on LN. I'm not sure if this would be too complicated to add.

examples/typescript/lnd/tsconfig.json
jamaljsr: I think it's fine to leave the comments. It's likely helpful for devs new to Typescript.

examples/typescript/lnd/src/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  forever(async () => {
+    const { action } = await inquirer.prompt<Action>({
+      type: 'list',
+      name: 'action',
+      message: `Hi ${info.alias}, what would you like to do?`,
+      choices: nodeHandler.getChoices(),
+    });
+    const answer = await nodeHandler.getAnswer(action);
+    console.log(answer);
+  });
```
jamaljsr, line 24: I'd suggest adding a try/catch around this to display a helpful error if there is an issue connecting to the node. `lnd`'s error messages can be difficult to digest for new devs.

jamaljsr (commented):
Awesome job so far! I like the direction you are headed with this. It was a great idea to create a console app to keep the code very simple.

I left a few comments just to give you my thoughts on what you have so far.

murtyjones:
> Awesome job so far! I like the direction you are headed with this. It was a great idea to create a console app to keep the code very simple.
> 
> I left a few comments just to give you my thoughts on what you have so far.

Thanks for reviewing so quickly @jamaljsr! I'm heading offline for a couple of weeks but will address your comments when back and notify when ready for re-review!

examples/typescript/lnd/src/nodeHandler.ts
```diff
@@ -0,0 +1,77 @@
...
+      case 'List open channels': {
+        const info = await this.lnrpc.listChannels();
+        return `Open channels:\n${JSON.stringify(info)}`;
+      }
+      default:
+        throw new Error('Question not recognized');
+    }
+  };
+
+  specifyChannelOpeningParameters = async () => {
```
murtyjones, line 58: @jamaljsr I haven't quite tidied the PR up, but I added the discuss command to open a channel with a peer. It seems to work fine, but I'm not familiar with lightning development so I don't know what I don't know. Could you take a look at this method and let me know if there are other things that you think need to be added or changed here?

examples/typescript/lnd/src/nodeHandler.ts
```diff
@@ -0,0 +1,77 @@
...
+    const { peers } = await this.lnrpc.listPeers();
+    const { pubKey } = await inquirer.prompt<{ pubKey: string }>({
+      type: 'list',
+      name: 'pubKey',
+      message: 'Which peer would you like to open a channel with?',
+      choices: peers.map(e => e.pubKey),
```
murtyjones, line 64: try as I might, I couldn't figure out how to list peers by alias (`getNodeInfo` doesn't seem to work for peers that aren't in your "graph" which I think means those without a valid payment path). Open to any feedback on how to display aliases

examples/typescript/lnd/src/nodeHandler.ts
jamaljsr: This looks good to me 👍

examples/typescript/lnd/src/nodeHandler.ts
jamaljsr: You are correct. It's not possible to `getNodeInfo` for nodes that are not in your graph. Unfortunately, there isn't anything we can do about that. One suggestion is you could catch the error and use an ellipsed pubkey `02d3...f422` instead of the alias for nodes not in the graph.

murtyjones:
@jamaljsr Finally got around to cleanup this up and it should be good for a re-review. LMK if you see anything to improve!

jamaljsr:
Thanks for the update. I'll try this out tomorrow and let you know what I think.

jamaljsr (approved):
Hey @murtyjones apologies for letting this sit for so long. 

The code looks great! I don't see much more that I would add. Thanks so much for contributing! 🥇

PR #419, opened 2021-01-18
feat(bitcoind): expose REST interface by default

valentinewallace:

### Description
Exposes bitcoind's REST interface by default. No prob if this idea has been considered then discarded for any reason, feel free to close. Thank you!

### Steps to Test

Attempt to connect to bitcoind over REST on startup.


jamaljsr:
Cool. I was not even aware that `bitcoind` had a REST interface. 

No objections from me on this one. I'll test it locally and merge if I don't run into any issues.

jamaljsr:
Tested and it works as advertised. Thanks for the update 🥇

PR #430, opened 2021-02-22
feat(channels): add private channel option

### LINKED ISSUE #425: Feature Request: Open private channels
For testing purposes it would be awesome to easily create private channels in polar

When opening a channel in polar it would be cool if there was a simple checkbox named "private" which should be unchecked by default.
Private channel lines could be displayed semitransparent in the view graph.

**Additional context**
I am testing stuff where I need to know if routes over private channels are found.

bolatovumar:
Closes #425

### Description

Added option to create a private channel. Private channels show up as low-opacity lines connecting nodes.

### Steps to Test

1. Create a test network
2. Open "Create a channel" modal
3. Notice there is now a "Make channel private" option
4. Check the aforementioned option
5. Notice that a private channel is created which shows up as a low-opacity line as opposed to a full-opacity line for regular channels

### Screenshots

Screen Shot 2021-02-21 at 6 25 17 PM


src/components/designer/custom/Link.tsx
```diff
@@ -88,6 +88,16 @@ const CustomLink: React.FC<ILinkDefaultProps> = ({
     };
   }, [link.properties, theme]);
 
+  const opacity = (() => {
```
bolatovumar, line 91: Create a lower-opacity line here if channel is private.

src/components/designer/lightning/actions/OpenChannelModal.tsx
```diff
@@ -60,13 +60,25 @@ const OpenChannelModal: React.FC<Props> = ({ network }) => {
     }
   }, [selectedFrom, selectedSats, nodes, sameNode, openChanAsync.loading]);
 
-  const handleSubmit = (values: any) => {
+  const handleSubmit = (values: {
```
bolatovumar, line 63: Replaced `any` with an inline type.

src/components/designer/lightning/actions/OpenChannelModal.tsx
```diff
@@ -60,13 +60,25 @@ const OpenChannelModal: React.FC<Props> = ({ network }) => {
...
     const toNode = lightning.find(n => n.name === values.to);
     if (!fromNode || !toNode) return;
     const autoFund = showDeposit && values.autoFund;
-    openChanAsync.execute({ from: fromNode, to: toNode, sats: values.sats, autoFund });
+    openChanAsync.execute({
+      from: fromNode,
+      to: toNode,
+      sats: values.sats,
+      autoFund,
+      isPrivate: values.isPrivate,
```
bolatovumar, line 80: Notice the new `isPrivate` option here.

src/components/designer/lightning/actions/OpenChannelModal.tsx
```diff
@@ -126,6 +138,9 @@ const OpenChannelModal: React.FC<Props> = ({ network }) => {
           </Checkbox>
         </Form.Item>
       )}
+      <Form.Item name="isPrivate" valuePropName="checked">
```
bolatovumar, line 141: New checkbox to make the channel private.

src/components/designer/link/Channel.tsx
```diff
@@ -42,6 +43,7 @@ const Channel: React.FC<Props> = ({ link, from, to }) => {
         />
       ),
     },
+    { label: l('isPrivate'), value: isPrivate ? 'true' : 'false' },
```
bolatovumar, line 46: I'm not sure if showing a `true` or `false` string here is the best. Maybe showing `Yes` or `No` would be better?

src/i18n/locales/de-DE.json
```diff
@@ -104,6 +104,7 @@
   "cmps.designer.link.Channel.capacity": "Kapazität",
   "cmps.designer.link.Channel.sourceBalance": "Quellensaldo",
   "cmps.designer.link.Channel.destinationBalance": "Zielguthaben",
+  "cmps.designer.link.Channel.isPrivate": "Is private",
```
bolatovumar, line 107: Not sure what to do with non-English locales.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -75,6 +75,7 @@ class CLightningService implements LightningService {
             localBalance: this.toSats(c.msatoshiToUs),
             remoteBalance: this.toSats(c.msatoshiTotal - c.msatoshiToUs),
             status,
+            isPrivate: c.private,
```
bolatovumar, line 78: Notice that I named the field `isPrivate` instead of a simpler `private` because `private` is a reserved keyword in JavaScript so I wanted to avoid it.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -107,11 +108,12 @@ class CLightningService implements LightningService {
     }
   }
 
-  async openChannel(
-    from: LightningNode,
-    toRpcUrl: string,
-    amount: string,
-  ): Promise<PLN.LightningNodeChannelPoint> {
+  async openChannel({
```
bolatovumar, line 111: Instead of adding yet another parameter to the function signature I replaced it with a single `options` parameter which contains all options which will be used to open a channel.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -122,6 +124,7 @@ class CLightningService implements LightningService {
       id: toPubKey,
       satoshis: amount,
       feeRate: '253perkw', // min relay fee for bitcoind
+      announce: !isPrivate,
```
bolatovumar, line 127: `c-lightning` uses the `announce` field to mark a channel as private. However, when you retrieve channels from `c-lightning` the channel is not marked as private (`private` field is always `false`).

src/lib/lightning/eclair/eclairService.spec.ts
```diff
@@ -73,6 +73,7 @@ describe('EclairService', () => {
       state: ELN.ChannelState.NORMAL,
       data: {
         commitments: {
+          channelFlags: 1,
```
bolatovumar, line 76: Added this because I updated the type from `any` to a more specific one so I had to add the missing property to the test. Channel flag `1` in `eclair` is for public channels.

src/lib/lightning/eclair/eclairService.ts
```diff
@@ -81,10 +81,11 @@ class EclairService implements LightningService {
           uniqueId: c.channelId.slice(-12),
           channelPoint: c.channelId,
           pubkey: c.nodeId,
-          capacity: commitInput.amountSatoshis,
+          capacity: commitInput.amountSatoshis.toString(10),
```
bolatovumar, line 84: Previously we were passing `capacity` as a number here when we expect a `string`.

src/lib/lightning/eclair/eclairService.ts
```diff
@@ -81,10 +81,11 @@ class EclairService implements LightningService {
...
           localBalance: this.toSats(localCommit.spec.toLocal),
           remoteBalance: this.toSats(localCommit.spec.toRemote),
           status,
+          isPrivate: c.data.commitments.channelFlags === 0,
```
bolatovumar, line 88: Channel flag `0` is for private channels in `eclair`.

src/lib/lightning/eclair/types.ts
```diff
@@ -33,13 +33,39 @@ export enum ChannelState {
...
+}
+
 export interface ChannelResponse {
   nodeId: string;
   channelId: string;
   state: ChannelState;
-  // there's a ton of data under this key that isn't needed, so just
-  // use any to avoid having to maintain the structure for each release
-  data: any;
+  data: ChannelData;
```
bolatovumar, line 68: Update the type from `any` to something more specific which include the properties we actually use.

src/lib/lightning/eclair/types.ts
```diff
@@ -54,7 +80,7 @@ export interface OpenChannelRequest {
   fundingSatoshis: number;
   pushMsat?: number;
   fundingFeerateSatByte?: number;
-  channelFlags?: number;
+  channelFlags?: ChannelFlags;
```
bolatovumar, line 83: Updated the type to a more specific one as the only possible values here are `0` and `1`.

src/lib/lightning/lnd/mappers.ts
```diff
@@ -25,4 +26,5 @@ export const mapPendingChannel = (status: LightningNodeChannel['status']) => (
   localBalance: chan.localBalance,
   remoteBalance: chan.remoteBalance,
   status,
+  isPrivate: false,
```
bolatovumar, line 29: Since `private` field doesn't exist on `PendingChannel` type so I'm setting it to `false` here. I guess I could also make the field nullable instead and assign it to `null` here?

src/utils/chart.ts
```diff
@@ -170,21 +171,24 @@ const updateLinksAndPorts = (
     properties: { nodeId: toNode.id },
   };
 
+  const properties: LinkProperties = {
```
bolatovumar, line 174: Just moved this into its own constant so I can specify the type and added the `isPrivate` field.

jamaljsr:
Hey @bolatovumar thanks for working on this. I'll work on reviewing this later today 👍

src/i18n/locales/de-DE.json
jamaljsr: You only need to update the `en-US.json` file. All of the other language files will be re-generated and overwritten by Crowdin.

Also, I'd prefer to capitalize the `P` to match the rest of the title case labels displayed in the sidebar.

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Thanks for the rationale 👍

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Good refactor. Thanks 🥇

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Do you know if the root cause of this problem is with the `c-lightning-REST` API or with `lightningd`? We should open an issue in their repos if the RPC isn't working as expected.

I don't think it's a good idea to display the "Make private" checkbox and then show a non-private channel after it's created. That makes it seem like Polar is broken.

src/lib/lightning/eclair/eclairService.spec.ts
jamaljsr: Great improvement. I should have done this initially instead of using `any`. 🥇

src/lib/lightning/eclair/eclairService.ts
jamaljsr: Good catch!

src/lib/lightning/lnd/mappers.ts
jamaljsr: Using `false` is fine with me here.

jamaljsr (changes requested):
Excellent PR @bolatovumar. The code looks really really good. Thank you for the refactoring improvements.

I just have two requests blocking approval:
1. The private channels for `c-lightning` nodes are not working correctly. I left a comment below on this. 
2. When you run `yarn test:ci`, you'll see there are still some conditionals not covered by test cases. I think you just need to add a couple more tests to cover the opposite case for `isPrivate`.

Also, just a suggestion in general, not a deal-breaker for this PR. It would be better to split up the changes into multiple smaller commits. This one is pretty large. I know it's easier to do it all in one go, but the smaller commits are much easier to review and result in a cleaner git history.

src/components/designer/link/Channel.tsx
bolatovumar: Ok, I will leave it as is then.

src/lib/lightning/clightning/clightningService.ts
bolatovumar: I don't know where the issue lies. I will have to look deeper into it. I believe I am using the right field here yet the channel is not private no matter what value I send it.

> I don't think it's a good idea to display the "Make private" checkbox and then show a non-private channel after it's created. That makes it seem like Polar is broken.
Yeah, I agree. I just wanted to create a PR anyway to start the discussion and figure out what's wrong in the process.

src/lib/lightning/lnd/mappers.ts
bolatovumar: Cool.

src/components/designer/custom/Link.tsx
bolatovumar: Done here: 60b888bf568a58ebe5e4de517a76157e27883d96.

src/i18n/locales/de-DE.json
bolatovumar: Updated capitalization here: 05b62b2a35c15538a0831ca2398b07a2c1de6d30

bolatovumar:
> 1. The private channels for `c-lightning` nodes are not working correctly. I left a comment below on this.
Yeah, I will look into this.

> 2. When you run `yarn test:ci`, you'll see there are still some conditionals not covered by test cases. I think you just need to add a couple more tests to cover the opposite case for `isPrivate`.
Will do.

> Also, just a suggestion in general, not a deal-breaker for this PR. It would be better to split up the changes into multiple smaller commits. This one is pretty large. I know it's easier to do it all in one go, but the smaller commits are much easier to review and result in a cleaner git history.
I assumed that the commits would be squashed into a single one at merge time for some reason so just did one giant commit. I would have split it into a bunch of smaller commits otherwise.

src/components/designer/custom/Link.tsx
jamaljsr: Hey, sorry to nitpick, but what I meant was to return `opacity` in the existing `useMemo` call that's already present on line 50. Since this block is already converting `link.properties` values into component vars, it makes sense to determine the `opacity` in there as well.

Something like this is what I had in mind:
```js
  const { leftStop, rightStop, leftColor, rightColor, opacity } = useMemo(() => {
    ...
    // the opacity of the link's path defaults to 100%
    let opacity = 1;
    if (link.properties) {
      const {
        type,
        direction,
        toBalance,
        capacity,
        isPrivate,
      } = link.properties as LinkProperties;
      if (type === 'open-channel') {
        ...
      }
      // use a lighter opacity for private channels
      if (isPrivate) opacity = 0.5;
    }
    return {
      leftStop,
      rightStop,
      leftColor,
      rightColor,
      opacity,
    };
  }, [link.properties, theme]);
```

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Ok cool, let me know if you'd like me to help troubleshoot in any way.

bolatovumar:
@jamaljsr I think I found where the bug lies with `c-lightning`. It's in the REST API implementation on this line: https://github.com/Ride-The-Lightning/c-lightning-REST/blob/22bb6c34960d09b8ec3c834a606b863446a3ef4f/controllers/channel.js#L77
Since this line double-bangs the string input for the `announce` field if it's `'0'` or `'false'` the result will always be `true` as double-bang of any non-empty string will be true, so the channel is never set to private. I will file the issue in that repo.

bolatovumar:
Opened an issue here: https://github.com/Ride-The-Lightning/c-lightning-REST/issues/55#issue-819545464

jamaljsr:
@bolatovumar Good find! 👍  

I'd like to keep this PR blocked until that issue is resolved in `c-lightning-REST` repo. I'll try to keep an eye on it, but feel free to update me here when that one gets resolved.

bolatovumar:
@jamaljsr I created a PR to fix the bug in the `c-lightning-REST` repo but I was wondering how I could test if it would actually fix the issue before that PR is merged. I tried changing this line: https://github.com/jamaljsr/polar/blob/f9a88ce022c9b2633209b31a5f6fee10518486cb/docker/clightning/Dockerfile#L137 to point to the tip of my for of the `c-lightning-REST` repo which has the fix but it doesn't seem to cut it. Do I need to build `c-lightning` from source using this Dockerfile or something?

jamaljsr:
That should work. Just edit the `Dockerfile`, then build the image. Polar should just use the new image. Did you use the same label `polarlightning/clightning:0.9.3`? Maybe try deleting the existing image before rebuilding. Also try adding `console.log`s to your fork repo to confirm Polar is using the new image.

bolatovumar:
> That should work. Just edit the `Dockerfile`, then build the image. Polar should just use the new image. Did you use the same label `polarlightning/clightning:0.9.3`? Maybe try deleting the existing image before rebuilding. Also try adding `console.log`s to your fork repo to confirm Polar is using the new image.

Ok, I build the Docker image while pointing to my fork of the `c-lightning-REST` repo which has this fix and after waiting for a while for it to build on my laptop it worked. See proof of screenshot:

Screen Shot 2021-03-05 at 10 16 56 PM

Will this fix get applied automatically once my PR in `c-lightning-REST` is merged in or will the images have to be re-built manually for all versions of `c-lightning`?

jamaljsr:
Thanks for the update. Lookin good 👍 

> Will this fix get applied automatically once my PR in `c-lightning-REST` is merged in or will the images have to be re-built manually for all versions of `c-lightning`?

Yes, the images would need to be rebuilt and pushed to DockerHub. Unfortunately, even after I do that, anyone who has already pulled the images from DockerHub will not automatically get the update. Only users who have never pulled the `c-lightning` image would receive the updated one when the network is started.

This is the first time where we've needed to update existing images. I'll have to put some thought into developing a strategy to handle this now as well as going forward. 

Off the top of my head, I have two ideas:
1. Create new image labels containing a suffix (ex: `polarlightning/clightning:0.9.3-1`). Then update the [node.json](https://github.com/jamaljsr/polar/blob/master/docker/nodes.json) file to include this suffix. We would just need to be mindful to maintain backwards compatibility with older versions of Polar.
2. Add a button on the **Manage Images** screen to remove local `polarlightning/*` images. The next time a network is started, it will pull the updated images. The downside is that this is not very obvious to users. They would not assume to do this when the `c-lightning` private channels are not displaying correctly.

If you have any feedback or other suggestions, please do share.

bolatovumar:
> 1. Create new image labels containing a suffix (ex: `polarlightning/clightning:0.9.3-1`). Then update the [node.json](https://github.com/jamaljsr/polar/blob/master/docker/nodes.json) file to include this suffix. We would just need to be mindful to maintain backwards compatibility with older versions of Polar.

In this case previous versions of `c-lightning` will continue working but will not get this fix, right? If that's the case I wonder if it makes sense to add a warning of some sort in the UI if a user is trying to open a private channel using an older version of the `c-lightning` image. It's a bit cumbersome but could be helpful.

> 2. Add a button on the **Manage Images** screen to remove local `polarlightning/*` images. The next time a network is started, it will pull the updated images. The downside is that this is not very obvious to users. They would not assume to do this when the `c-lightning` private channels are not displaying correctly.

This might still be useful even if it's not obvious as there could be other use cases for this if a user wants to remove local images for some other unrelated reason in the future.

bolatovumar:
Related `c-lightning-REST` PR merged: https://github.com/Ride-The-Lightning/c-lightning-REST/pull/56

jamaljsr:
Thanks for the update. I'll update the images soon.

Regarding the UI changes. I like your idea about showing the warning when opening a private channel. We could use the docker image hash to determine if the user has the new image or an old one pulled locally. If they have the old one, instruct them to click the button on the **Manage Images** screen. This seems like a reasonable solution. Thanks for the suggestion 👍

jaonoctus (approved):

jamaljsr:
I rebased this PR and made a couple fixes and improvements. Thanks again for adding this feature. Apologies for taking so long to merge.

PR #550, opened 2022-04-16
Update custom-nodes.md

niteshbalusu11:
Change golang version in Dockerfile when building custom node.


jamaljsr (approved):
Thank for the update! 👍

PR #572, opened 2022-07-22
feat(images): add support for CLN v0.11.2

ktecho:
### Description

Adds support for CLN v0.11.2. I've built the image locally and it seems to work fine.

This should not be merged until v0.11.2 docker image is generated.

There is another PR to update nodejs in the docker image to v14 (LTS) because v12 seems to be unsupported. Then you can build manually the new image as stated in that other PR.


jamaljsr:
Hey @ktecho thanks so much for the recent PRs. I am currently away and will check them out next week. Just wanted to give you the heads up immediately.

jamaljsr:
> There is another PR to update nodejs in the docker image to v14 (LTS) because v12 seems to be unsupported. Then you can build manually the new image as stated in that other PR.

I'm curious why you didn't include the nodejs upgrade in this PR?

ktecho:
> > There is another PR to update nodejs in the docker image to v14 (LTS) because v12 seems to be unsupported. Then you can build manually the new image as stated in that other PR.
> 
> I'm curious why you didn't include the nodejs upgrade in this PR?

You're probably right. I was thinking that first you need to create the docker image and push it to the docker registry. If you  merge this and the docker image is not in the registry, it would fail for people running master.

jamaljsr:
Yeah, I think it would be better to include it in this PR. When I test the changes, I'll push the image to Docker Hub before merging.

Thanks again for contributing.

ktecho:
Perfect. I'll do it later tonight. Thanks!

ktecho:
@jamaljsr you should now be able to create the CLN v0.11.2 image with the newer nodejs and then try it with just this PR.

jamaljsr:
Thanks for the update. I'll review this by tomorrow. If all looks good, I'll push the image to DockerHub and merge this PR.

jamaljsr (commented):
I tested locally and didn't run into any issues. I pushed the new image to DockerHub.

One last change that needs to be done in order for the new version to be detected by existing Polar users is to bump the `version` field to 38 in the `nodes.json` and `constants.ts` files.

jamaljsr (approved):
Awesome! Thanks for the contribution and updates.

> What is that number for?

This is so Polar can easily determine if the version of the `nodes.json` file that is saved locally is different from the one fetched from Github when checking for updates (see [repoService.ts](https://github.com/jamaljsr/polar/blob/master/src/lib/docker/repoService.ts#L53)).

PR #581, opened 2022-08-06
feat(nodes): better visually know what node is selected

ktecho:
### Description

This gives the user the ability to better see which node is selected. Before, a small shadow would highlight the selected node, but it's difficult to see with the light theme and almost impossible to see with the dark theme, at least in my monitor.

### Screenshots

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "history": "4.10.1",
     "husky": "8.0.1",
     "i18next": "21.9.0",
-    "i18next-browser-languagedetector": "6.1.4",
+    "i18next-browser-languagedetector": "6.1.5",
     "i18next-scanner": "4.0.0",
     "jest-environment-jsdom-sixteen": "1.0.3",
     "js-yaml": "4.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1768,13 +1768,6 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.14.6":
-  version "7.14.6"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.14.6.tgz#535203bc0892efc7dec60bdc27b2ecf6e409062d"
-  integrity sha512-/PCB2uJ7oM44tz8YhC4Z/6PeOKXp4K588f+5M3clr1M4zbqztlo0XEfJ2LEzj/FgwfgGcIdl8n7YYjTCI0BYwg==
-  dependencies:
-    regenerator-runtime "^0.13.4"
-
 "@babel/runtime@^7.15.4":
   version "7.15.4"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.15.4.tgz#fd17d16bfdf878e6dd02d19753a39fa8a8d9c84a"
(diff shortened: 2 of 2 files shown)


jamaljsr:
Hey Luis, thanks for another contribution. Appreciate the screenshots as well.

While I agree with the intent of this PR, I'm not really a fan of how the border looks on the node. It feels like a bit too much contrast now.

What do you think about brightening up the shadow by changing the color to orange? It should stand out more while still being somewhat subtle.

ktecho:
You're right in that it's probably too aggresive, but I can't even match my clothes, so who knows :)

I don't like this either, but how do you see it?

image

image

This ones has bigger shadow:

image

image

jamaljsr:
Personally, I prefer the smaller shadow. I think that's enough distinction to easily identify the currently selected node.

ktecho:
@jamaljsr changed. Please try to see if it's not so ugly.

Also, thanks for pointing me on how to get the `isSelected` variable from the state.

src/theme/colors.ts
```diff
@@ -9,7 +9,7 @@ export interface ThemeColors {
   statusTag: { stopped: string };
   statusBadge: { default: string };
   link: { default: string };
-  node: { background: string; border: string };
+  node: { background: string; border: string; borderSelected: string };
```
jamaljsr, line 12: The changes in this file appear to be left over from your first attempt with the borders. Can you discard these since they aren't used anymore?

src/components/designer/custom/NodeInner.tsx
```diff
@@ -6,10 +6,12 @@ import { ThemeColors } from 'theme/colors';
...
 import { Loader, StatusBadge } from 'components/common';
 import NodeContextMenu from '../NodeContextMenu';
+import { useStoreState } from 'store';
 
 const Styled = {
-  Node: styled.div<{ size?: ISize; colors: ThemeColors['node'] }>`
-    border: 1px solid ${({ colors }) => colors.border};
+  Node: styled.div<{ size?: ISize; colors: ThemeColors['node']; isSelected: boolean }>`
+    ${props =>
+      props.isSelected ? 'box-shadow: 2px 2px 5px rgb(255 147 40 / 40%)' : null};
```
jamaljsr, line 14: While testing, I started tweaking the box shadow a bit to see if I could improve it at all. I like the look more with the value `box-shadow: 5px 5px 20px rgb(255 147 40 / 40%)`. What do you think?

jamaljsr (commented):
Thanks for the PR. I left some comments below.

src/theme/colors.ts
ktecho: You're right. I forgot about hose. Deleting...

src/components/designer/custom/NodeInner.tsx
ktecho: Works for me.

ktecho:
@jamaljsr Changes done. Please give it a look :)

jamaljsr (changes requested):
Cool. I tested it and the code looks good.

One final request from me is to rebase on master and squash the 5 commits into a single one. We don't need to have many commits for such a small change.

ktecho:
The changes were implemented in a separate PR:  https://github.com/jamaljsr/polar/pull/583

PR #583, opened 2022-08-12
style(nodes): better visually know what node is selected

ktecho:
### Description

This gives the user the ability to better see which node is selected. Before, a small shadow would highlight the selected node, but it's difficult to see with the light theme and almost impossible to see with the dark theme, at least in my monitor.


ktecho:
This PR is related with https://github.com/jamaljsr/polar/pull/581. Go there for more details.

It seems somehow the git workflow got screwed while trying to squash commits, probably because I'm not acustomed to the VS Code IDE.

jamaljsr (approved):
tACK LGTM 🥇 

Thanks for the contribution.

PR #584, opened 2022-08-15
fix(lightningd): display CA cert and distinguish it from client cert and key

louneskmt:
### Description

This PR moves `TLS Cert` and `TLS Key` to `TLS Client Cert` and `TLS Client Key` in order to avoid confusion with the server certificate+key or the CA certificate+key. It changes `TLS Cert` to the `ca.pem` certificate instead of the `client.pem` certificate.

Translations are also included for the new strings.

### Steps to Test

1. Create a new network
2. Add a Core Lightning Node
3. Go to the Connect tab
4. You can check the displayed certificates are identical with the files certificates:
```shell
$ xxd -p <path-to-polar-network>/volumes/c-lightning/<node name>/lightningd/regtest/ca.pem | tr -d '\n'
$ xxd -p <path-to-polar-network>/volumes/c-lightning/<node name>/lightningd/regtest/client.pem | tr -d '\n'
$ xxd -p <path-to-polar-network>/volumes/c-lightning/<node name>/lightningd/regtest/client-key.pem | tr -d '\n'
```

### Screenshots

imageimageimage


jamaljsr (changes requested):
Tested and works as advertised. 🎉

Great job on the initial changes being really spot on. The code looks good, the unit tests are passing, and the commit messages follow the standard.

The only change I see would be for the translations. You only need to modify the `en-US.json` file in your PR. The other locale files are auto-generated by Crowdin after PRs get merged into the master branch. See #579. Please discard the changes to all of the other locale files, then this PR will be good to go.

Thanks so much for the contribution!

louneskmt:
Will the changes be applied to networks previously created? During my tests, it worked great for a new network but an old one was just displaying the TLS Cert and Admin Macaroon. This is due to the paths name in the network file no being the same as before: it was looking for `paths.tlsCert`, `paths.tlsClient` and `paths.tlsClientKey` but found `paths.tlsCert` and `paths.tlsKey` in the file (generated on a previous version).

jamaljsr:
Unfortunately, this wouldn't affect existing networks. The paths are generated when the network is created so only new networks would show these values in the sidebar.

louneskmt:
Couldn't we regenerate some parts of the`networks` file on start? These paths are deterministic so we could rerun it safely, right?

jamaljsr:
Yes, it's possible. If you think it's worthwhile to implement, you could add a migration to update the cert paths for CLN nodes. I've done this in the past as you can see in the [migrations.ts](https://github.com/jamaljsr/polar/blob/master/src/utils/migrations.ts#L139) file. That should give you good examples of how to implement this behavior.

louneskmt:
Done ✅ . AFAIU, this requires tagging a new version `v1.4.0`, but I don't think it is a big issue, as Polar definitely needs to update its binaries (the last release was more than a year ago). WDYT?

jamaljsr:
Thanks for the migration update. I just tested it and it works well.

Yes, I am definitely planning to tag a new release before the end of this month. There are just a couple small features I'd like to squeeze in before shipping it.

jamaljsr:
Ah! I forgot to check the test coverage. It looks like this migration scenario isn't tested. Can you add a test in the [dockerService.spec.ts](https://github.com/jamaljsr/polar/blob/master/src/lib/docker/dockerService.spec.ts#L458) file.

PR #607, opened 2022-10-10
Update custom-nodes.md

hieblmi:
Docker `--no-cache` flag for fresh build.

### DIFF
--- a/docs/custom-nodes.md
+++ b/docs/custom-nodes.md
@@ -101,7 +101,7 @@ Polar v1.0.0 supports using custom docker images for nodes in your networks. The
 
    ```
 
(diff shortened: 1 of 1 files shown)


jamaljsr:
Thanks for the contribution. Can you explain why you want to add this flag? What problem is this solving?

hieblmi:
Yes apologies - I saw that the docker build results were cached hence additional changes to the referenced lnd-repo were ignored. By specifying the `--no-cache` flag docker fetched the changes.

jamaljsr:
Ah gotcha. That makes sense. Thanks for the fix 👍

PR #643, opened 2023-01-02
feat(network-add-taro): adds parameter to createnetwork

amovfx:
helper function to make taro nodes.

### Description

[Description of your changes go here. Please provide both casual and technical explanations.]
A proposal to make a testing network with taro nodes. I noticed that a testRepoState was available that provided easy access to change node versions.


amovfx:
Hi, this is a proposal I made last night that I thought would do the trick for creating test network ith Taro. I got the idea from the testRepoState you had in the helper file. It's easy to adjust, Maybe the dockerRepoState should be parameterized? What do you think?

PR #645, opened 2023-01-02
test(taro): fixed up tests

amovfx:
A few tests were failing when checking the amount of latest nodes on the default side bar.

### Description

Tests pass.

### Steps to Test

Run tests


jamaljsr:
I'm aware that the tests are not passing and that there are many more tests that need to be written to cover all of the new Taro related code. This is on the TODO list in #641 and something I will address after I squash a few more of the open issues.

I'd prefer there not be a bunch of small PRs targeting the Taro branch. instead, just create one PR per big feature.

PR #646, opened 2023-01-02
Feat/network-add-taro-with-tests

amovfx:
### Description
createnetwork helper function is parameterized with taro option.
tests updated


amovfx:
New to pull req, sorry if this has been a little bit rocky. If you don't like this approach, which I made previously, I now understand what you were talking about your recommendation of adding the lnd and taro nodes explicitly per test. I'm down to do that if you think it's best.

jamaljsr:
As I mentioned in https://github.com/jamaljsr/polar/pull/641#issuecomment-1368512949 I'd say we shouldn't add taro nodes when creating the initial network. Going to close this PR.

PR #661, opened 2023-01-24
Feat/new-address-modal

amovfx:
Hey Jamal,
This is new-address-modal for your to review the inital work flow.

I still have to do tests and some polish on the code to bring it into line with what I learned on the last pr, but it's functional and ready for critique.


amovfx:
Oh man I think I messed up the rebase, might want to hold off.

Just to clarify, I should rebase this branch onto feat/add-taro correct?

jamaljsr:
> Just to clarify, I should rebase this branch onto feat/add-taro correct?

Yes, that is correct. You should drop any commits that are already in that branch.

amovfx:
This has been a very good learning experience.
This is good to review preliminary workflow. If you want to review the code, that would be cool, but I plan on doing a pass over to get into line with what I've learned from your previous notes.

jamaljsr:
Thanks for the updates. I'll do an initial UX review by tomorrow. 

Thanks for continuing to work on these updates.

amovfx:
You're welcome. Thanks for taking the time to work with me on this.

jamaljsr:
Overall it works good on the UX side. I do have some suggestions for improvement.

1. I don't think it's a good idea to instantly generate a new address every time one of the input fields changes. This is going to cause many addresses to be created in the taro db. For example, if I change the amount field from `1` to `1000`, it will create 3 addresses, 2 of which I do not need. These addresses are persisted by taro and can be viewed via `tarocli addrs query`. 
2. Rename "Import from Taro Node" to "Choose asset from Taro node". The term "Import" is a bit confusing.
3. Instead of displaying the address in the initial form, I think it would be better to keep the flow similar to how invoices are created. Initially display the fields that the user can modify. When they click the Generate button, the form is hidden and the address is displayed along with a "Copy & Close" button.
4. When you copy the address, the message that pops up at the top of the screen shows the full address and looks pretty bad. This should be ellipsed so it's not so wide.
5. In the Actions sidebar, the "New Address" button should be full width, not half width.

amovfx:
Excellent feedback, this is working way better. If you are happy with this, I will do a final pass on the code to attempt to bring it up to the standard of your previous reviews.

jamaljsr:
Just tested https://github.com/jamaljsr/polar/pull/661/commits/77f2756a1d4599fdcfccdca33007dad108af004d. This feels much better. Just a couple more comments about the latest updates.
1. After clicking Generate and displaying the address, the modal's footer should no longer be displayed. It allows the user to click Generate button again, but it doesn't actually do anything.
2. After choosing a token from another node, the Amount field is updated to contain the full amount. I'm not sure most people would want to drain the sending node's wallet in one payment. I think it'll be more common to want to send a bunch of smaller payments. So maybe it makes more sense to set this to a small number, like `10`, and don't change it when an asset is selected. 
 
On a side note, as I've been playing more with generating addresses, I've noticed a pain point. When minting a token I keep forgetting to change the amount. I realize it when I go to create an address. Then I need to go back and create a new asset. I think we should increase the default amount in the Mint Asset modal to `1000`.

amovfx:
This is really shaping up. Thanks for the feedback.

amovfx:
This is ready for review.
Something that bothers me, I don't really have a way to quickly validate a taro address off the top of my head. 
I think it would be cool to have a regex function that checks to make sure that a genesis info is valid. Are all taro addresses the same length?

jamaljsr:
Cool. I'll go through the code tomorrow.

Regarding the data, I honestly think its better if Polar is lenient when it comes to validation. If we get too strict and tightly coupled to the implementations, it increases the chances of breaking changes with future node releases. I prefer to minimize the need to ship a new Polar release just to support a new version of a node. I am fine with not validating the data user's enter. We just need to display the error that the nodes will respond with. It also gives devs more flexibility for testing without Polar getting in the way.

src/components/designer/taro/actions/NewAddressButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+interface Props {
+  node: TaroNode;
+  type?: 'button' | 'menu';
+}
+
+const OpenSendAssetModal: React.FC<Props> = ({ node, type }) => {
+  const { l } = usePrefixedTranslation('cmps.designer.taro.actions.NewAddressButton');
+  const { showNewAddress } = useStoreActions(s => s.modals);
+  const btnlabel = l('newAddress');
```
jamaljsr, line 22: This is super nitpicky but I can be pretty anal about consistency in a code base. :)

I'd prefer to keep the syntax identical to the other Button components. There isn't much benefit to declaring this var here versus placing the value in the JSX below.

src/store/models/taro.ts
```diff
@@ -123,10 +129,19 @@ const taroModel: TaroModel = {
...
       });
-
       return res;
     },
   ),
+  getNewAddress: thunk(
+    async (actions, { node, genesisBootstrapInfo, amount }, { injections }) => {
+      const api = injections.taroFactory.getService(node);
+      const address = await api.newAddress(node, {
+        genesisBootstrapInfo: Buffer.from(genesisBootstrapInfo as string, 'hex'),
```
jamaljsr, line 139: The ` as string` can be removed since the type is already set to `string` on `NewAddressPayload`

jamaljsr (commented):
Took a first pass on the code. It's mostly all good. Just some suggestions to simplify the components.

Also, be sure to run `yarn lint` to catch the couple linter warnings.

electron/tarod/tarodProxyServer.ts
```diff
@@ -59,6 +59,15 @@ const mintAsset = async (args: {
   return await rpc.mintAssets(args.req);
 };
 
+const newAddress = async (args: {
+  node: TarodNode;
+  req: TARO.NewAddressRequest;
+}): Promise<TARO.NewAddressResponse> => {
+  const rpc = await getRpc(args.node);
+  debug('newAddress req', JSON.stringify(args, null, 2));
```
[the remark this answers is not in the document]
amovfx: done

src/components/designer/taro/actions/NewAddressButton.tsx
amovfx: done

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+import { useStoreActions, useStoreState } from 'store';
+import { NewAddressPayload } from 'store/models/taro';
+import { Network } from 'types';
+import { ellipseInner } from 'utils/strings';
+import { format } from 'utils/units';
+import CopyableInput from 'components/common/form/CopyableInput';
+import TaroAssetSelect from 'components/common/form/TaroAssetSelect';
+import TaroNodeSelect from 'components/common/form/TaroNodeSelect';
+
+const TaroBalanceSelect = TaroAssetSelect;
```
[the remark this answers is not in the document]
amovfx: Fixed with collapse and taro select refactor.

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+  useEffect(() => syncChart(network), []);
+
+  const handleSelectedBalance = (value: number) => {
+    if (balances) {
+      setSelectedBalance(balances[value]);
+      setSelectedGenesisBootstrapInfo(balances[value].genesisBootstrapInfo);
+      setSelectedAmount(balances[value].balance);
+      form.setFieldsValue({
+        genesisBootstrapInfo: balances[value].genesisBootstrapInfo,
+        amount: Math.min(parseInt(balances[value].balance), 10),
```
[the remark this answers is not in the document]
amovfx: done

src/i18n/locales/en-US.json
```diff
@@ -293,6 +296,22 @@
...
+  "cmps.designer.taro.actions.NewAddressModal.import": "Choose asset from Taro node",
+  "cmps.designer.taro.actions.NewAddressModal.amount": "Amount",
+  "cmps.designer.taro.actions.NewAddressModal.amountPlaceholder": "1",
+  "cmps.designer.taro.actions.NewAddressModal.address": "Address",
+  "cmps.designer.taro.actions.NewAddressModal.addressPlaceholder": "Address",
+  "cmps.designer.taro.actions.NewAddressModal.okBtn": "Generate",
+  "cmps.designer.taro.actions.NewAddressModal.cancelBtn": "Cancel",
+  "cmps.designer.taro.actions.NewAddressModal.submitError": "Error:",
+  "cmps.designer.taro.actions.NewAddressModal.successTitle": "Successfully created address",
+  "cmps.designer.taro.actions.NewAddressModal.successDesc": "{{assetName}} {{amount}}",
```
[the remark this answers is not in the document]
amovfx: Done

src/store/models/taro.ts
amovfx: done

src/components/designer/taro/actions/NewAddressButton.tsx
amovfx: done.

src/components/common/form/TaroNodeSelect.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export interface Props extends SelectProps<SelectValue> {
+  name: string;
+  label?: string;
+  help?: string;
+  nodeStatus?: Status;
+  networkNodes?: CommonNode[];
+}
+
+const TaroNodeSelect: React.FC<Props> = ({
```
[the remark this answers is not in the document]
amovfx: Done, but I think its going to def need a look over, it is a pretty big change.

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+  Spacer: styled.div`
+    height: 12px;
+  `,
+};
+
+interface Props {
+  network: Network;
+}
+
+const NewAddressModal: React.FC<Props> = ({ network }) => {
```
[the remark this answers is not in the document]
amovfx: How about dumping the collapse all together and just having the TaroAssetSelect with options?

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const [isMenuCollapsed, setIsMenuCollapsed] = useState(true);
+
+  //component local variables
+  const [form] = Form.useForm();
+  const thisTaroNode = network.nodes.taro.find(
+    node => node.name === nodeName,
+  ) as TarodNode;
+  const otherTaroNodes = network.nodes.taro.filter(node => node.name !== nodeName);
+
+  useEffect(() => syncChart(network), []);
```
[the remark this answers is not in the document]
amovfx: Done.

amovfx:
I ended up doing a pretty big refactor. The collapse panel is gone, I thought it was a bit redundant, but if you think it's a good idea to hide the select balance functionality behind it, I'll put it back.

src/components/common/form/TaroDataSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const node: TaroNodeModel = nodes[taroNetworkNode.name];
+      if (node) {
+        const data: TaroModelData[] = node[resourceKey] as TaroModelData[];
+        if (data) {
+          return {
+            label: taroNetworkNode.name,
+            options: data.map((taroData: TaroModelData, j: number) => ({
+              label: taroData.name,
+              value: `${taroData.name}-${i}`,
+              key: cantorPairing(i, j),
```
jamaljsr, line 56: This `cantorPairing` function is pretty cryptic. It seems to me that its simply calculating a unique number to use for the `key` prop. Any reason why a simple `${i}-$[j}` wouldn't work?

Also, there is no `key` prop on the `Option` according to the [docs](https://ant.design/components/select#option-props), so you should be able to just take this out.

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,185 @@
...
+            />
+          </Form.Item>
+          <Styled.Spacer />
+
+          <Styled.TaroDataSelect
+            name="node"
+            label={l('selectBalance')}
+            taroNetworkNodes={otherTaroNodes}
+            selectBalances
+            onChange={v => handleSelectedBalance(v?.valueOf() as PTARO.TaroBalance)}
```
jamaljsr, line 128: Using `valueOf()` is a bit of a code smell in Typescript, similar to using `any`. We should try to avoid it unless absolutely necessary. I would suggest updating the `TaroDataSelect.onChange` prop to use the proper type, then you wouldn't need to do this type casting.

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,185 @@
...
+          />
+          <Styled.Spacer />
+          <Form.Item label={l('amount')} name="amount">
+            <InputNumber
+              onChange={v => setSelectedAmount(v?.valueOf()?.toString() as string)}
+              min={'1'}
+              disabled={selectedBalance?.type === PTARO.TARO_ASSET_TYPE.COLLECTIBLE}
+            />
```
jamaljsr, line 136: There's a few things that could be improved here. Take a look at how `<InputNumber>` is used throughout the code base and try to mirror those.

For example, here's how it's implemented in the `OpenChannelModal` component:
```ts
        <InputNumber<number>
          disabled={openChanAsync.loading}
          formatter={v => `${v}`.replace(/\B(?=(\d{3})+(?!\d))/g, ',')}
          parser={v => parseInt(`${v}`.replace(/(undefined|,*)/g, ''))}
          style={{ width: '100%' }}
          onChange={v => setSelectedSats(v as number)}
        />
```

src/components/designer/taro/actions/NewAddressModal.tsx
```diff
@@ -0,0 +1,185 @@
...
+          amount: format(`${selectedAmount}`),
+        })}
+        extra={
+          <Form>
+            <Form.Item>
+              <CopyableInput label="Invoice" value={taroAddress} />
+            </Form.Item>
+            <Form.Item>
+              <CopyToClipboard text={taroAddress} onCopy={handleCopy}>
+                <Button type="primary">{l('copyClose')}</Button>
```
jamaljsr, line 157: THis should be "Address" instead of "Invoice".

jamaljsr (commented):
This is definitely a step in the right direction. The `NewAddressModal` component looks much better. I have more follow-up comments based on your latest changes.

src/components/designer/taro/actions/NewAddressModal.tsx
amovfx: Done

src/components/common/form/TaroDataSelect.tsx
amovfx: Done, I was getting a lot of key errors when I was first building this, they are gone now.

src/components/designer/taro/actions/NewAddressModal.tsx
amovfx: Done

src/components/designer/taro/actions/NewAddressModal.tsx
amovfx: Done. Yea my typescript game is weak. Lots to learn here.

src/components/designer/taro/actions/NewAddressModal.tsx
amovfx: Done

src/components/designer/taro/actions/MintAssetButton.tsx
```diff
@@ -8,7 +8,7 @@ import { useStoreActions } from 'store';
 
 const Styled = {
   Button: styled(Button)`
-    width: 50%;
+    width: 100%;
```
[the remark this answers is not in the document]
amovfx: Done

src/components/designer/taro/actions/NewAddressModal.tsx
amovfx: I seem to be using this in everything that I have done so far. What do you think of me moving this out to its own component in a refactor? I'm thinking something like components/common/form/AmountInput.tsx?

jamaljsr (approved):
Alright. This looks good to go now. Thanks for iterating on this and getting the feature complete. 👌

PR #668, opened 2023-02-08
fix(network-card): added taro nodes stat to network card

amovfx:
### Description

Updated Network card to display the amount of taro nodes.

### Steps to Test

1. Create a new network with taro nodes.
2. Go to network selection panel.


amovfx (commented):
Reviewed with tests and ci passing.

src/components/home/NetworkCard.tsx
```diff
@@ -43,6 +43,13 @@ const NetworkCard: React.FC<{ network: Network }> = ({ network }) => {
             suffix={<LinkOutlined />}
           />
         </Col>
+        <Col span={12}>
```
jamaljsr, line 46: I'm not a big fan of the empty space in the second column. You could get all 3 on one row by setting `span={8}` on all columns and removing the "Nodes" from the titles.

src/components/home/NetworkCard.tsx
```diff
@@ -43,6 +43,13 @@ const NetworkCard: React.FC<{ network: Network }> = ({ network }) => {
...
+          <Statistic
+            title={l('taroNodes')}
+            value={network.nodes.taro.length}
+            suffix={<ApiOutlined />}
```
jamaljsr, line 50: antd doesn't provide a lot of icons to choose from but I think the `DollarOutlined` one makes more sense for Taro since the primary goal of the project is to bring stablecoins to Bitcoin.

jamaljsr (commented):
Great idea! I didn't even think to update these. Thank you.

Just have a few minor suggestions.

src/components/home/NetworkCard.tsx
amovfx: Done.

jamaljsr (approved):
This looks good to go. Going to merge. Thanks for the update

PR #672, opened 2023-02-13
fix(bad buffer): mint asset metadata

amovfx:
Metadata needed to be a base64 string.
Should I write some tests for the asset info drawer? 

### Description

Meta Data was being encoded wrong. This is fixed.

### Steps to Test

1. Mint an asset
2. Add anything meta data field
3. Open AssetInfoDrawer. You should see meta data at top
### Screenshots

[Only if applicable]


jamaljsr (approved):
tACK. Looks good. Thanks for the fix.

PR #683, opened 2023-02-25
Fix/filter-empty-taro-nodes

amovfx:
Taro nodes are now filtered if they dont have any assets or balances.


jamaljsr (approved):
tACK. Thanks for the update.

PR #685, opened 2023-02-26
fix(general bugs): workflow fixes

amovfx:
-taro nodes connect in order when dropped into the graph
-Taro info asset draw title says 'No Assets' if there are no assets

### DIFF
--- a/src/components/designer/taro/info/AssetsList.tsx
+++ b/src/components/designer/taro/info/AssetsList.tsx
@@ -11,7 +11,7 @@ import AssetInfoDrawer from './AssetInfoDrawer';
 
 const Styled = {
   Wrapper: styled.div``,
-  P: styled.p`
+  Empty: styled.p`
     text-align: center;
   `,
 };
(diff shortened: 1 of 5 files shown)


jamaljsr (commented):
For the Info sidebar, I could have been more specific. The problem I see with this is having a divider with no content below it looks weird. So I was thinking to add text below the divider stating "This node does not have any assets.", so there is content below it.

It looks like code coverage for the `network.ts` file dropped with this update.

amovfx:
Thanks for that clarity. Sorry for missing that code coverage. Ill update this asap.

amovfx:
```
 network.ts                           |   99.69 |      100 |   98.86 |     100 |
```
I haven't seen this before, I have 100% line coverage, why would statements and functions be less than 100?

jamaljsr:
Most likely you have an `if` statement but the `else` condition is never hit. Open the `coverage/lcov-report/index.html` for more details on what's not covered.

amovfx:
Super interesting how a nuanced a language can be. Looks like a sorted array apparently wont sort if it doesn't need to. Pretty cool optimization.

src/components/designer/taro/info/AssetsList.tsx
```diff
@@ -10,6 +11,9 @@ import AssetInfoDrawer from './AssetInfoDrawer';
 
 const Styled = {
   Wrapper: styled.div``,
+  P: styled.p`
```
jamaljsr, line 14: This should have a more descriptive name, such as `Empty`.

jamaljsr (commented):
tACK 👍

This looks good now. Just one small nitpick.

src/components/designer/taro/info/AssetsList.tsx
amovfx: Done.

jamaljsr (approved):

PR #691, opened 2023-03-18
Update golang version of custom docker image

hieblmi:
1.18-alpine -> 1.19.7-alpine


jamaljsr:
Thanks for the update 👌

PR #703, opened 2023-04-03
fix(chart.ts): pending channel display bug

### LINKED ISSUE #699: Bug: Source and destination of unconfirmed channel is shown reverse
**Describe the bug**
When I open a channel (through cli) between two lnd nodes the channel details (sidebar) show source and destination in reverse. Only after the channel open transaction is mined, the source and destination are correct.

**To Reproduce**
1. Launch terminal for `alice`
2. Run `lncli openchannel <bob-pubkey> 100000`
3. Check channel details
4. Source will be `bob` and destination `alice`
6. Mine 6 blocks
7. Source and destination will be the reverse

**Expected behavior**
Source should be `alice` and destination `bob`

- OS: Arch Linux
- Polar v1.4.0
- Docker v23.0.1
- Compose v2.16.0

**Additional context**
- only between two lnd nodes (v0.15.1-beta)
- only while funding tx is unconfirmed

amovfx:
Channel info now displays proper source and dest names
for pending channels

Closes #699

### Description

Channel links are proccessed twice resulting in data being clobbered.

### Steps to Test

1. Launch terminal for alice
2. Run ```lncli openchannel <bob-pubkey> 100000```
3. Check channel details

### Screenshots

[Only if applicable]

### DIFF
--- a/src/lib/lightning/lnd/lndService.ts
+++ b/src/lib/lightning/lnd/lndService.ts
@@ -45,13 +45,27 @@ class LndService implements LightningService {
     } = await proxy.pendingChannels(this.cast(node));
 
     const pluckChan = (c: any) => c.channel as LND.PendingChannel;
+    const isChanInitiatorLocal = (c: LND.PendingChannel) =>
+      c.initiator === LND.Initiator.INITIATOR_LOCAL;
     // merge all of the channel types into one array
     return [
       ...open.filter(c => c.initiator).map(mapOpenChannel),
-      ...opening.map(pluckChan).map(mapPendingChannel('Opening')),
-      ...closing.map(pluckChan).map(mapPendingChannel('Closing')),
-      ...forceClosing.map(pluckChan).map(mapPendingChannel('Force Closing')),
-      ...waitingClose.map(pluckChan).map(mapPendingChannel('Waiting to Close')),
+      ...opening
+        .map(pluckChan)
+        .filter(isChanInitiatorLocal)
+        .map(mapPendingChannel('Opening')),
+      ...closing
+        .map(pluckChan)
+        .filter(isChanInitiatorLocal)
+        .map(mapPendingChannel('Closing')),
+      ...forceClosing
+        .map(pluckChan)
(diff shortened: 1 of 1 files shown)


amovfx:
@jamaljsr 
You might want to take a look at this a bit more closely, it was subtle. I'm also curious if taro is going to need this. I don't think so because the links can't really be changed after the network starts.

src/lib/lightning/lnd/lndService.ts
```diff
@@ -45,13 +45,27 @@ class LndService implements LightningService {
...
+    const isChanInitiatorLocal = (c: LND.PendingChannel | LND.Channel) =>
+      c.initiator === LND.Initiator.INITIATOR_LOCAL;
     // merge all of the channel types into one array
     return [
-      ...open.filter(c => c.initiator).map(mapOpenChannel),
-      ...opening.map(pluckChan).map(mapPendingChannel('Opening')),
-      ...closing.map(pluckChan).map(mapPendingChannel('Closing')),
-      ...forceClosing.map(pluckChan).map(mapPendingChannel('Force Closing')),
-      ...waitingClose.map(pluckChan).map(mapPendingChannel('Waiting to Close')),
+      ...open.filter(isChanInitiatorLocal).map(mapOpenChannel),
```
jamaljsr, line 52: THis won't work because `LND.Channel.initiator` is a `boolean`, but the `isChanInitiatorLocal` function is comparing it to an `enum`/`number`. That function looks good for all of the pending arrays below, but the open channels array has a different type so it should be filtered differently.

amovfx: Ah ty!

jamaljsr:
The code looks good. I'll test it when I get a chance.

Can you squash your commits into one and use a proper message prefix `fix(lnd): ...`

jamaljsr:
Tested and confirmed this resolves the issue. Just needs a squash then this can be merged.

amovfx:
Fantastic, thanks for the guidance and feedback once again.

PR #774, opened 2023-09-01
feat(images): add support for Core Lightning v23.08

### LINKED ISSUE #772: Feature Request: Add support for CLN v23.08
Would it be possible to add support for CLN v23.08?

jamaljsr:
Closes #772 

Adds support for Core Lightning v23.08.

**⚠️ Draft Status**

This PR is not yet fully functional due to the latest release breaking compatibility with the [c-lightning-REST](https://github.com/Ride-The-Lightning/c-lightning-REST) plugin which Polar uses. In Core Lightning v23.08, a new `clnrest` plugin was added and enabled by default. When both of these plugins are enabled, `lightningd` will fail to start with the error:
```
2023-09-01T04:12:05.631Z DEBUG   plugin-manager: started(35) /opt/c-lightning-rest/plugin.js
error starting plugin '/usr/local/libexec/c-lightning/plugins/clnrest/clnrest.py': option name 'rest-protocol' is already taken
```

There is a short-term workaround which is to disable the included `clnrest` plugin using the flag `--disable-plugin=clnrest.py`, then the `c-lightning-REST` plugin will function as it has previously. This will require a new release of Polar in order to make sure this flag is only passed to CLN nodes v23.08 and higher. The downside is that devs using Polar won't be able to use the new REST support.

The long-term solution would be to update Polar to use the new `clnrest` plugin for it's communication with the `lightningd` API. I want Polar to continue to support older CLN nodes, so it would still have to be able to use the `c-lightning-REST` API until `v23.05.2` and below are completely phased out. I try to support the last 3-4 more major versions of node implementations, so this could take a year or more. Until that happens, Polar will need to support both REST APIs side-by-side. 

I am opening this PR now since some of the work is done. The docker images are pushed to Docker Hub and I am able to get the new CLN release to run in Polar as long as I manually add the `--disable-plugin=clnrest.py` flag via Advanced Options. It's going to take a bit more work to decide on a path forward then implement the changes necessary.


toneloc:
Am interested in support for Core Lightning v24.02 or v24.02.1.

This version has support for dual-funded channels.  

Any roadmap for that? Thanks.

jamaljsr:
I plan to circle back to this PR pretty soon to try to get CLN working properly again. Unfortunately, I can't give an exact timeline, but it may be in the order of weeks because a new release of Polar may be needed to support the latest CLN version.

toneloc:
Thanks. I think dual-funded channels would look awesome in Polar. 

But yeah first things first, and happy to test when available and I will check in on #863 too.

When CLN started to include the REST API by default, it broke some things for me too.

jamaljsr:
This PR has been replaced by #879.

PR #795, opened 2023-10-19
fix download broken links + update download links to v2.1.0

ramigs:
fixing broken download links for Mac and Windows.

also noticed that a new version was released yesterday, so updated the links to point to `v2.1.0`. 
pls let me know if keeping `v2.0.0` is intended and I'll update the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,11 @@ You will be prompted to install Docker if Polar cannot detect it automatically
 
 ## Download
 
-Download Polar v2.0.0 for your OS
+Download Polar v2.1.0 for your OS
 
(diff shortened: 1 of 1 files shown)


jamaljsr (approved):
Thanks so much for catching this. I did forget to update the README with the latest release.

ramigs:
btw, to avoid having to manually update at each release, has it been considered linking to the `latest` release?

or is it unsuitable because some releases are `rc` and therefore not to be linked?
if so, maybe could be considered introducing tag `stable` and linking to it.

however, this change would require file names to not have the version anymore (not sure if it's a hard requirement)

PR #826, opened 2023-12-31
feat: add xdg support

uwla:
## Description

If and only if `~/.local/share/polar` exists, Polar will use it to store application data instead of `~/.polar`, abiding to the XDG specification (see https://wiki.archlinux.org/title/XDG_user_directories).

This is a backwards compatible update: users who have their data at `~/.polar` will NOT be affected. The `~/.local/share/polar` XDG-compliant directory is not created automatically and the user must create it manually to enable the new feature.

## Note

I tried to import `dataPath` from `src/utils/config` in `src/shared/utils` to avoid code duplication but it wouldn't compile and would complain `cannot resolve module 'utils/config'`. I tried to use a relative import, `import { dataPath } from '../utils/config'`, but the compiler would throw similar errors (it would not resolve `shared/<filename>` imports inside `../utils/config`). Therefore, there is some duplicated code.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -3,6 +3,7 @@ import log from 'electron-log';
 import { join } from 'path';
 import http from 'http';
 import https from 'https';
+import { existsSync } from 'fs';
 
 /**
  * setup logging to store log files in ~/.polar/logs/ dir
@@ -10,7 +11,10 @@ import https from 'https';
 export const initLogger = () => {
   log.transports.file.resolvePath = (variables: log.PathVariables) => {
     const ap = app || remote.app;
-    return join(ap.getPath('home'), '.polar', 'logs', variables.fileName as string);
+    const home = ap.getPath('home');
+    const xdgPath = join(home, '.local', 'share', 'polar');
+    const dataPath = existsSync(xdgPath) ? xdgPath : join(home, '.polar');
+    return join(dataPath, 'logs', variables.fileName as string);
   };
 };
 
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -3,11 +3,19 @@ import { join } from 'path';
 import { NodeImplementation } from 'shared/types';
 import { Network } from 'types';
 import { dockerConfigs } from './constants';
+import { existsSync } from 'fs';
+
+/**
+ * XDG-compliant path where application data is stored
+ */
+export const xdgDataPath = join(remote.app.getPath('home'), '.local', 'share', 'polar');
 
 /**
  * root path where application data is stored
  */
-export const dataPath = join(remote.app.getPath('home'), '.polar');
+export const dataPath = existsSync(xdgDataPath)
+  ? xdgDataPath
+  : join(remote.app.getPath('home'), '.polar');
 
 /**
  * legacy path where application data was stored in v0.1.0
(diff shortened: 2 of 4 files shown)


uwla:
I compiled Polar with the feature of PR and started using it for local development of LN applications. While using Polar compiled from the master branch, I experienced some errors when creating multiple channels. At first, I thought this was caused by my patch but it is not. The most up-to-date versions (take the last two or four commits) of the official master branch have the same errors: I compiled them and ran the App Image to test.

So, I created a path of this PR (`git diff HEAD~1 > patch`), switched to the latest stable version (`git checkout v2.0`) and built Polar (`yarn package`). It has been working perfectly so far. Therefore, I am confident my PR does not introduce any bug and the new functionality won't affect users unless they explicitly create the `~/.local/share/polar/` directory.

src/shared/utils.ts
```diff
@@ -3,14 +3,18 @@ import log from 'electron-log';
...
 
 /**
  * setup logging to store log files in ~/.polar/logs/ dir
  */
 export const initLogger = () => {
   log.transports.file.resolvePath = (variables: log.PathVariables) => {
     const ap = app || remote.app;
-    return join(ap.getPath('home'), '.polar', 'logs', variables.fileName as string);
+    const home = ap.getPath('home');
+    const xdgPath = join(home, '.local/share/polar');
```
jamaljsr, line 15: I haven't tested but I believe this will fail on Windows because it uses the `\` path separator. 

Use `join(home, '.local', 'share', 'polar')` to make it compatible with all platforms.

src/utils/config.ts
```diff
@@ -3,11 +3,19 @@ import { join } from 'path';
 import { NodeImplementation } from 'shared/types';
 import { Network } from 'types';
 import { dockerConfigs } from './constants';
+import { existsSync } from 'fs';
+
+/**
+ * XDG-compliant path where application data is stored
+ */
+export const xdgDataPath = join(remote.app.getPath('home'), '.local/share/polar');
```
jamaljsr, line 11: Same as above

src/utils/config.ts
```diff
@@ -3,11 +3,19 @@ import { join } from 'path';
...
 
 /**
  * root path where application data is stored
  */
-export const dataPath = join(remote.app.getPath('home'), '.polar');
+export const dataPath = existsSync(xdgDataPath)
+  ? xdgDataPath
+  : join(remote.app.getPath('home'), 'polar');
```
jamaljsr, line 18: Shouldn't this be `.polar` instead of `polar`?

jamaljsr (commented):
Thanks for the PR. This is a nice update, especially since it's backwards compatible.

I just requested some minor fixes then this should be good to go.

uwla:
@jamaljsr I have addressed the changes.

jamaljsr (approved):
tACK LGTM 🚀

Thanks for the fixes.

jamaljsr (changes requested):
Ah, I just realized the test coverage dropped. Wouldn't you mind adding a test to cover the missing line?

uwla:
> Ah, I just realized the test coverage dropped. Wouldn't you mind adding a test to cover the missing line?

I'll do it later on.

uwla:
@jamaljsr, tests added.

jamaljsr (approved):
Thanks for adding the tests. This looks good to go 🚀

PR #827, opened 2024-01-05
Added support for Node.js version 20

kelvinator07:
Closes #(820)

### Description
Added support for Node.js version 20


package.json
```diff
@@ -9,8 +9,11 @@
     "url": "https://lightningpolar.com"
   },
   "main": "public/main/electron/index.js",
+  "engines": {
+    "node": ">=20.10.0"
+  },
```
jamaljsr, line 14: I think this should be set to the minimum version required for Polar to run. In my testing, v20.0.0 works but anything before that doesn't. I'd suggest changing this to `>=20.10.0`.

package.json
```diff
@@ -79,9 +83,9 @@
     "@types/jest": "29.5.11",
     "@types/js-yaml": "4.0.9",
     "@types/node": "18.17.19",
-    "@types/react": "17.0.66",
+    "@types/react": "18.2.46",
     "@types/react-copy-to-clipboard": "5.0.7",
-    "@types/react-dom": "18.2.8",
+    "@types/react-dom": "18.2.18",
```
jamaljsr, line 88: Do we have to bump the react types? Since we are still using v17, I'm concerned about the types not matching the actual JS code being executed and causing weird errors.

src/components/network/NetworkActions.tsx
```diff
@@ -96,7 +96,7 @@ const NetworkActions: React.FC<Props> = ({
 
   const mineAsync = useMiningAsync(network);
 
-  const handleClick: MenuProps['onClick'] = useCallback(info => {
+  const handleClick: MenuProps['onClick'] = useCallback((info: { key: any }) => {
```
jamaljsr, line 99: nit: try to minimize the use of `any` when possible. Using `string` should work fine here.

jamaljsr (commented):
Thanks so much for this PR. This update was long overdue.

package.json
kelvinator07: You Mean >=20 ?

jamaljsr: lol.. yes. That is what I meant to write.

package.json
kelvinator07: Yes We Have To Bump It Up. It's Still Compatible With v17. Bumping The Type To v18 Would Require Alot Of Changes In The Code And Would Also Require Bumping Up Other Dependencies Like Electron.. e.t.c
Meanwhile, Did You Encounter Any Weird Errors While Testing/Running The App?

jamaljsr: I haven't run into any issues in my testing. This just makes me a bit nervous, but I guess it should be fine. I definitely don't think we should go down that rabbit hole of upgrading React & Electron. That is going to be a painful project which I hope to get to one day.

jamaljsr (approved):
I tested the latest changes. LGTM 👌

Thanks for this long awaited fix.

PR #828, opened 2024-01-16
Fix: Bug: Eclair - Error in GUI when open channel #804

### LINKED ISSUE #804: Bug: Eclair - Error in GUI when open channel
**Describe the bug**
When I open a channel within 2 Eclair nodes I get the error Unable to get node information from xxx (eg Alice, Bob) in GUI - in the terminal I can see the channel and can work with it, but in the GUI I can't see it and the GUI subsequently not working properly

**To Reproduce**
1. Create new network -> 2 Eclair v0.9.0 + 2 Bitcoin Core v25.0
2. Start network
3. Deposit funds (at Alice) - At node go to "Actions" and click on "Deposit" for 1,000,000 sats
4. Open channel to Bob (either via "Actions" -> "Open Channel" - "Ourgoing", or drag "from dot to dot" within the graphic display of the network). There is no need to change anything in the settings for opening a new node
5. Then I see:
image
  subsequently in the node view (on each of them) I see error:
image
   but in terminal channel exist:
`eclair@alice:/app$ eclair-cli allchannels`
[
  {
    "shortChannelId": "108x1x1",
    "a": "03210acf8ff1d593478ffc948fa3461e219824972ca3878949486fc3a990c27b0a",
    "b": "03c302b2f55a4385dbdf8157baff4f25b471b7f1e5566df714e9438e5e2ed6bfd0"
  }
]

- OS: Microsoft Windows 10 Home; version 10.0.19045 N/A Build 19045
- Polar Version: v2.1.0
- Docker Version: v20.10.21
- Docker Compose Version: v2.13.0

kelvinator07:
This PR Closes #804

Some [ChannelData](https://github.com/jamaljsr/polar/blob/master/src/lib/lightning/eclair/types.ts#L46) properties for the Eclair node was updated for v0.9.0 when calling its [API](https://acinq.github.io/eclair/#channels-2) and this caused a bug when running that version. 

This PR updates the ChannelData properties to be compatible with v0.9.0, v0.8.0 and v0.7.0.

### DIFF
--- a/src/lib/lightning/eclair/eclairService.spec.ts
+++ b/src/lib/lightning/eclair/eclairService.spec.ts
@@ -74,20 +74,28 @@ describe('EclairService', () => {
       state: ELN.ChannelState.NORMAL,
       data: {
         commitments: {
-          channelFlags: 1,
-          localParams: {
-            isFunder: true,
-            isInitiator: undefined as any,
-          },
-          localCommit: {
-            spec: {
-              toLocal: 100000000,
-              toRemote: 50000000,
+          params: {
+            localParams: {
+              isFunder: true,
+              isInitiator: undefined as any,
+            },
+            channelFlags: {
+              announceChannel: true,
             },
           },
-          commitInput: {
-            amountSatoshis: 150000,
-          },
+          active: [
+            {
+              fundingTx: {
+                amountSatoshis: 150000,
+              },
+              localCommit: {
+                spec: {
+                  toLocal: 100000000,
+                  toRemote: 50000000,
+                },
+              },
+            },
+          ],
         },
       },
     };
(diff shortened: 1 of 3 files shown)


jamaljsr:
Thanks for tackling this issue. I have a bit of feedback after testing this locally.

1. This is unrelated to the code and more about housekeeping. It would be great if you could write more in the PR description explaining the rationale behind the changes you've made. What was the root of the problem? Did something change in the daemons or dependencies that caused this issue. This would just be helpful for reviewers, as well as our future selves to understand why specific code was changed.
2. It looks like this works with the latest version of Eclair, but adding older versions to the network is now broken with similar errors. I guess they changed the structure of the API responses with the latest version. It would be great if our code could support both the old and new structures. If it's too much of a pain, then the only other option would be to drop support for the older versions. I prefer to keep the last 2 major versions around for testing, so that would be a last resort.

kelvinator07:
> Thanks for tackling this issue. I have a bit of feedback after testing this locally.
> 
> 1. This is unrelated to the code and more about housekeeping. It would be great if you could write more in the PR description explaining the rationale behind the changes you've made. What was the root of the problem? Did something change in the daemons or dependencies that caused this issue. This would just be helpful for reviewers, as well as our future selves to understand why specific code was changed.
> 2. It looks like this works with the latest version of Eclair, but adding older versions to the network is now broken with similar errors. I guess they changed the structure of the API responses with the latest version. It would be great if our code could support both the old and new structures. If it's too much of a pain, then the only other option would be to drop support for the older versions. I prefer to keep the last 2 major versions around for testing, so that would be a last resort.

Thanks for the feedback. I've added a description to the PR.
The [API doc](https://acinq.github.io/eclair/#channels-2) doesn't have the sample responses for older versions. Had to run each versions while logging its responses to get that data. I've updated the PR to be compatible with the last 3 versions. Do take a look and let me know what you think. Thanks

jamaljsr:
Thanks for the updates. I will test it out this wknd.

jamaljsr (approved):
tACK LGTM 🚀

Thanks for the changes. This looks great! I tested all 3 versions and was able to open channels and make direct payments. 

There seems to be an issue with routing payments through Eclair v8.0+ nodes, but that is out of scope for this PR.

Issue #418, opened 2021-01-16
Feature Request: Implement real-time updates of channel balances from the nodes


murtyjones:
**Describe the bug**

While working on an [example app](https://github.com/murtyjones/polar/tree/feat/add-channel-opening-example) using Polar, I noticed that opening a channel in the app doesn't seem to reflect in the Polar UI. I assume this is a bug but not 100% sure if Polar is intended to show channels in the pending state.

**To Reproduce**
1. Create a network in Polar (one node for each implementation should be fine), deposit some funds for Alice, and ensure that her balance is confirmed
1. Run the example app linked above (`yarn && yarn start` in the example app directory should do it)
1. Open a channel with one of the peers using the prompts (ignore the fact that pubkeys are being used instead of Aliases for now 😅 )
1. Execute the `List pending channels` command and you should see that you have one

**Expected behavior**
I would think that Polar would reflect that there's a pending channel.

**Screenshots**
![image](https://user-images.githubusercontent.com/708562/104826998-6f58ee00-5827-11eb-9402-d078b80ddf01.png)
![image](https://user-images.githubusercontent.com/708562/104827050-d8d8fc80-5827-11eb-8371-b497dc184a29.png)

**Desktop:**
- OS: Catalina
- Polar Version: Master
- Docker Version: 20
- Docker Compose Version: 1.27.4

jamaljsr:
This isn't a bug, more of a feature request. Polar currently doesn't automatically update the UI when events occur on the network outside of Polar. We need to implement subscriptions to the `lnd` streaming RPCs and maybe do polling on the other implementations that don't support streaming/websockets.

For now, to sync the UI with the state of the nodes, first click on the canvas, then click on the Sync icon at the top of the sidebar.

I'll keep this issue open and this is something I've wanted to implement but havent gotten around to it yet.

murtyjones:
> For now, to sync the UI with the state of the nodes, first click on the canvas, then click on the Sync icon at the top of the sidebar.

Totally missed the sync button, thank you!

murtyjones:
> We need to implement subscriptions to the `lnd` streaming RPCs

I've started to play around with this but dealing with an issue that I'm not familiar enough with Electron/this app's setup to resolve.

If we create an IPC event for establishing the stream:
```
subscribeChannelEvents: 'subscribe-channel-events',
...
const subscribeChannelEvents = async (args: { node: LndNode }) => {
  const rpc = await getRpc(args.node);
[...]

PR #837, opened 2024-02-27
Channel UI feature: Real-time updates of channels status

### LINKED ISSUE #603: Feature Request: Add websockets support for core-lightning
core-lightning introduced websockets to talk the node using `--experimental-websocket-port=PORT` flag. Their unix sockets is something very hard to get it working with docker. Maybe websockets should be easy to add?

Their gRPC is too incomplete to be able to perform any meaningful dev work and REST is managed in-house and uses external libraries which I am not a fan of. Maybe websockets could be the way to go until gRPC gets developed.

kelvinator07:
Closes [#603](https://github.com/jamaljsr/polar/issues/603) and [#418](https://github.com/jamaljsr/polar/issues/418)

### Description
Opening/Closing a lightning channel outside of Polar doesn't reflect on the UI. A click on the Sync button is required to update the UI.

The feature aims to update the UI automatically every time it detects a channel is opened, pending, or closed between lightning nodes. This is implemented by subscribing to each lightning node on startup via a listener that listens for channel events that occur outside of the Polar app and updates the UI accordingly.


jamaljsr (commented):
Thanks for working on this feature. I've wanted this for a pretty long time.

Regarding the functionality, it mostly works as expected. I just have these suggestions:
1. I tested by opening a Terminal on `alice` and running `lncli openchannel ...`. It didn't display the channel until I mined a few blocks. I know LND sends channel events as soon as the channel opening txn is broadcast. I think it should show pending channels in the UI when this happens.
2. I noticed the `syncChart` is called many times when I have a few LND nodes in the network. This function is pretty heavy since is makes a bunch of RPC calls for each node. It would be better to only run `syncChart` once, maybe it can be throttled by time.
3. Right now there's only listeners for channel open events. It would be nice to also listen for channel closes and sent/received payments. I think it would be good to have at least two because it'll force you to create better abstractions for the listeners.

I also left some specific feedback on the code below. I hope this is helpful.

kelvinator07:
@jamaljsr  Thanks for the feedback. Would update and revert.

electron/lnd/lndProxyServer.ts
```diff
@@ -117,6 +124,25 @@ const decodeInvoice = async (args: {
   return await rpc.decodePayReq(args.req);
 };
 
+const getChannelListener = async (args: { node: LndNode }): Promise<any> => {
+  try {
+    const responseChan = `lnd-${ipcChannels.getChannelListener}-response`;
+    const rpc = await getRpc(args.node);
+    const stream = rpc.subscribeChannelEvents();
+    stream.on('data', (data: LND.ChannelEventUpdate) => {
+      const dataAsString = JSON.stringify(data);
```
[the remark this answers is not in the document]
kelvinator07: Updated

electron/lnd/lndProxyServer.ts
```diff
@@ -4,6 +4,13 @@ import createLndRpc, * as LND from '@radar/lnrpc';
 import { ipcChannels, LndDefaultsKey, withLndDefaults } from '../../src/shared';
 import { LndNode } from '../../src/shared/types';
 
+let _webContents: Electron.WebContents;
+
+export const setupWebContents = (webContents: Electron.WebContents) => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  _webContents = webContents;
+};
```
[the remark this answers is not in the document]
kelvinator07: You are right, using a callback is better. Updated

src/lib/ipc/ipcService.ts
```diff
@@ -53,6 +57,11 @@ export const createIpcSender = (serviceName: string, prefix: string) => {
         JSON.stringify(stripNode(uniqPayload), null, 2),
       );
       ipcRenderer.send(reqChan, uniqPayload);
+      if (callback) {
+        ipcRenderer.on(resChan, (event, data) => {
```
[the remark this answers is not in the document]
kelvinator07: I've used the `resChan + node port` for uniqueness

src/lib/lightning/eclair/eclairService.ts
```diff
@@ -250,6 +252,20 @@ class EclairService implements LightningService {
...
 
+  async getChannelListener(node: LightningNode): Promise<any> {
+    const lndNode: EclairNode = this.cast(node);
+    const base64auth = Buffer.from(`:${eclairCredentials.pass}`).toString('base64');
+    const config = {
+      url: `127.0.0.1:${lndNode.ports.rest}`,
+      headers: {
+        Authorization: `Basic ${base64auth}`,
+      },
+    };
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/types/index.ts
```diff
@@ -155,6 +155,10 @@ export interface LightningService {
     invoice: string,
     amount?: number,
   ) => Promise<PLN.LightningNodePayReceipt>;
+  getChannelListener: (
+    node: LightningNode,
+    callback?: (data: string) => void,
+  ) => Promise<any>;
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/store/models/lightning.ts
```diff
@@ -290,6 +292,57 @@ const lightningModel: LightningModel = {
...
     },
   ),
+  addListeners: thunk(
+    async (actions, network, { injections, getStoreState, getStoreActions }) => {
+      const { nodes } = getStoreState().network.networkById(network.id);
+      for (const node of nodes.lightning) {
+        try {
+          await actions.getAllInfo(node);
+        } catch {}
+      }
```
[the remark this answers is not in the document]
kelvinator07: This should be in the Switch case. Updated

src/store/models/lightning.ts
```diff
@@ -290,6 +292,57 @@ const lightningModel: LightningModel = {
...
+
+      for (const node of nodes.lightning) {
+        switch (node.implementation) {
+          case 'LND':
+            await injections.lightningFactory
+              .getService(node)
+              .getChannelListener(node, async (message: string) => {
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/store/models/network.ts
```diff
@@ -780,7 +780,11 @@ const networkModel: NetworkModel = {
       // ensure that each node is aware of the entire graph and can route payments properly
       if (lnNodesOnline.length) {
         await Promise.all(lnNodesOnline)
-          .then(async () => await getStoreActions().lightning.connectAllPeers(network))
+          .then(async () => {
+            await getStoreActions().lightning.connectAllPeers(network);
+            // Add listeners to lightning nodes
+            await getStoreActions().lightning.addListeners(network);
```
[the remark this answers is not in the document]
kelvinator07: Yes, the listeners are added when the network is started, so it makes sense to remove them when the network is stopped. This has been implemented.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,9 +197,27 @@ class CLightningService implements LightningService {
     );
   }
 
+  async getChannelListener(node: LightningNode): Promise<any> {
+    const lndNode: CLightningNode = this.cast(node);
+    const config = {
+      url: `127.0.0.1:${lndNode.ports.rest}`,
+      macaroonPath: String(lndNode.paths.macaroon),
+      tls: String(lndNode.paths.tlsCert),
+    };
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/lib/lightning/lnd/lndService.ts
```diff
@@ -186,6 +186,21 @@ class LndService implements LightningService {
     );
   }
 
+  async addListenerToNode(node: LightningNode): Promise<void> {
```
jamaljsr, line 189: As I mentioned in a comment above, there really isn't a need to add/remove the listener for LND nodes. But you will need to keep these functions declared here in order to conform to the `LightningService` interface. They can just be empty functions that do not do anything in this case.

src/lib/lightning/eclair/eclairApi.ts
```diff
@@ -40,3 +43,21 @@ export const httpPost = async <T>(
...
+      Authorization: `Basic ${base64auth}`,
+    },
+  };
+  return config;
+};
+
+export const setupListener = (eclairNode: EclairNode): EclairWebSocket => {
+  const eclairConfig = setupConfig(eclairNode);
+  const eclairTs = new EclairTs(eclairConfig);
+  const listener = eclairTs.listen();
```
jamaljsr, line 61: I feel like adding the `eclair-ts` package just to call this one method is a bit much, especially since that package is no longer maintained. It hasn't been updated since Eclair v0.3.0. 

I would suggest to just implement the relevant code we need here in this file. You could just have a `listen` function that returns the WebSocket.

src/lib/lightning/eclair/eclairService.ts
```diff
@@ -250,6 +254,43 @@ class EclairService implements LightningService {
     );
   }
 
+  async addListenerToNode(node: LightningNode): Promise<void> {
+    this.listener = setupListener(this.cast(node));
```
jamaljsr, line 258: There's just a single listener here for all Eclair nodes. If I create and start a network with 3 Eclair nodes, then remove one of the nodes, it will stop listening on all Eclair nodes. We don't want this to happen. We'll probably need to keep a mapping of from node->listener and only close the stop listening on the node that gets removed.

jamaljsr (commented):
Thanks for the additional updates. This is looking better. I didn't get a change to review all of the code in the free time I had this morning, but here is some feedback on what I've gone through so far. I will continue reviewing the rest and doing more testing later this week.

src/store/models/lightning.ts
```diff
@@ -290,6 +293,108 @@ const lightningModel: LightningModel = {
       );
     },
   ),
+  addListeners: thunk(async (actions, network, { injections, getStoreState }) => {
+    const { nodes } = getStoreState().network.networkById(network.id);
+    nodes.lightning.forEach(
+      async node =>
+        await injections.lightningFactory.getService(node).addListenerToNode(node),
+    );
```
jamaljsr, line 301: Calling `await` in a `forEach` won't actually wait for the promises to resolve. Is this the behavior you intended? Right now, it is going to add the listeners while simultaneously subscribing to channel events. So it'll be a race to get there first and may produce unespected results. To actually make the execution wait for the promises to resolve, you can either use a `for/of` loop or you can use `await Promise.all(network.lightning.map(...))`

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,6 +199,43 @@ class CLightningService implements LightningService {
...
+  }
+
+  async subscribeChannelEvents(
+    node: LightningNode,
+    callback: (event: PLN.LightningNodeChannelEvent) => void,
+  ): Promise<void> {
+    // check c-lightning channels every 30 seconds
+    this.channelsInterval = setInterval(() => {
+      this.checkChannels(node, callback);
+    }, 3 * 10000);
```
jamaljsr, line 220: This is sharing one timer for all nodes, so only the last node started will receive events. Keep in mind that there will only ever be one instance of this service for the whole app, even when there are multiple nodes or multiple networks. So any state in here needs to be separated for each node.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,6 +199,43 @@ class CLightningService implements LightningService {
...
+    const channels = await this.getChannels(node);
+    channels.forEach(channel => {
+      if (channel.pending) {
+        callback({ type: 'Pending' });
+      } else if (channel.status === 'Open') {
+        callback({ type: 'Open' });
+      } else if (channel.status === 'Closed') {
+        callback({ type: 'Closed' });
+      }
+    });
```
jamaljsr, line 236: This is going to call the callback every 30secs for all channels the node has. We are trying to mimic the behavior of subscribing to channel events which will only notify us when something changes with the node's channels. If the channels all have the same status as the previous check, it shouldn't call the callback because nothing has changed.

src/utils/async.ts
```diff
@@ -49,3 +49,24 @@ export const waitFor = async (
     }, interval);
   });
 };
+
+let timer: NodeJS.Timeout | null = null;
+let immediateCall = true;
+let lastCallTime = 0;
+
+export const debounceFunction = async (func: () => Promise<void>) => {
```
jamaljsr, line 57: Is there a reason why we can't use the `debounce` fuction from `lodash` instead of writing our own? My concern would be the shared variables here being used by simultaneous calls to `debounceFunction`.

jamaljsr (commented):
I was able to get through the rest of the code. Here's some additional feedback.

electron/lnd/lndProxyServer.ts
```diff
@@ -117,6 +125,46 @@ const decodeInvoice = async (args: {
...
+  try {
+    if (lndRpc == null) {
+      await setupLndRpcForListener({ node: args.node });
+    }
+
+    const responseChan = `lnd-${ipcChannels.subscribeChannelEvents}-response-${args.node.ports.rest}`;
+    const stream = lndRpc?.subscribeChannelEvents();
+    if (stream) {
+      stream.on('data', (data: LND.ChannelEventUpdate) => {
+        const dataAsString = JSON.stringify(data);
```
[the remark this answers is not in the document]
kelvinator07: If we don't have a single format to handle all event types, we will have to create multiple handleEventCallbacks to handle all the different event types we will subscribe to in the future. For example Chanel events - LND.ChannelEvent format, balance event - LND.BalanceEvent format. e.t.c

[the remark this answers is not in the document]
kelvinator07: Using `any` makes sense then

electron/lnd/lndProxyServer.ts
```diff
@@ -117,6 +125,46 @@ const decodeInvoice = async (args: {
   return await rpc.decodePayReq(args.req);
 };
 
+let lndRpc: LND.LnRpc | null = null;
```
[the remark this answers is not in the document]
kelvinator07: Thanks for your explanation Jamal. This has been updated

electron/lnd/lndProxyServer.ts
```diff
@@ -4,6 +4,14 @@ import createLndRpc, * as LND from '@radar/lnrpc';
 import { ipcChannels, LndDefaultsKey, withLndDefaults } from '../../src/shared';
 import { LndNode } from '../../src/shared/types';
 
+let handleEventCallback: (responseChan: string, message: string) => void;
```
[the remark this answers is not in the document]
kelvinator07: Done

src/lib/ipc/ipcService.ts
```diff
@@ -18,14 +23,28 @@ const stripNode = (payload: any) => {
   return payload;
 };
 
+const formatResponse = (data: any): PLN.LightningNodeChannelEvent => {
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/lib/lightning/clightning/clightningService.spec.ts
```diff
@@ -200,24 +200,74 @@ describe('CLightningService', () => {
...
         'test-error',
       );
-      expect(clightningApiMock.httpGet).toBeCalledTimes(4);
+      expect(clightningApiMock.httpGet).toHaveBeenCalledTimes(4);
+    });
+  });
+
+  it('should add Listener To Node', async () => {
+    await clightningService.addListenerToNode(node);
+  });
```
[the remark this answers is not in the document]
kelvinator07: Removed

src/lib/lightning/eclair/eclairApi.ts
kelvinator07: Good suggestion. Replaced with only the function it needs

src/lib/lightning/eclair/eclairService.ts
kelvinator07: Updated

src/store/models/lightning.ts
```diff
@@ -290,6 +293,108 @@ const lightningModel: LightningModel = {
...
+      async node =>
+        await injections.lightningFactory.getService(node).removeListener(node),
+    );
+  }),
+  addChannelListeners: thunk(
+    async (actions, network, { injections, getStoreState, getStoreActions }) => {
+      const { nodes } = getStoreState().network.networkById(network.id);
+      for (const node of nodes.lightning) {
+        switch (node.implementation) {
+          case 'LND':
```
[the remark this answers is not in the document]
kelvinator07: Yes no difference, except for eclair that needs an extra call to `getAllInfo(node)`

src/store/models/lightning.ts
kelvinator07: True, no the behaviour I want. Updated to using a for/of

src/utils/async.ts
kelvinator07: This has been updated to use `lodash.throttle`. Throttle is the right function to use.

kelvinator07:
@jamaljsr This PR has been updated based on your last comments and is ready to be reviewed.

src/lib/lightning/eclair/eclairApi.ts
```diff
@@ -40,3 +42,53 @@ export const httpPost = async <T>(
...
+const setupConfig = (eclairNode: EclairNode): ELN.ConfigOptions => {
+  const base64auth = Buffer.from(`:${eclairCredentials.pass}`).toString('base64');
+  const config = {
+    url: `127.0.0.1:${eclairNode.ports.rest}`,
+    headers: {
+      Authorization: `Basic ${base64auth}`,
+    },
+  };
+  return config;
+};
```
jamaljsr, line 55: THis is great! Can we also use this in the `request` function to reduce duplication?

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -20,6 +20,8 @@ const ChannelStateToStatus: Record<CLN.ChannelState, PLN.LightningNodeChannel['s
   };
 
 class CLightningService implements LightningService {
+  channelsInterval: NodeJS.Timeout | null = null;
```
jamaljsr, line 23: It looks like this isn't used any longer and can be removed.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,9 +199,100 @@ class CLightningService implements LightningService {
...
+  ) => {
+    channels.forEach(channel => {
+      if (channel.pending) {
+        callback({ type: 'Pending' });
+      } else if (channel.status === 'Open') {
+        callback({ type: 'Open' });
+      } else if (channel.status === 'Closed') {
+        callback({ type: 'Closed' });
+      }
+    });
```
jamaljsr, line 272: This is going to call the `callback` multiple times. One for each channel, even if only one channel changed. The throttling will prevent syncing the chart more than once but from an API design perspective, this feels like unexpected behavior. The `subscribeChannelEvents` for other implementations only sends a message for the single channel that has been updated.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,9 +199,100 @@ class CLightningService implements LightningService {
...
+
+  private areChannelsEqual(
+    channels1: { pending: boolean; status: string }[],
+    channels2: { pending: boolean; status: string }[],
+  ): boolean {
+    return (
+      channels1.length === channels2.length &&
+      JSON.stringify(channels1) === JSON.stringify(channels2)
+    );
+  }
```
jamaljsr, line 287: This implementation could result in a false positive if the order of the channels are not the same in subsequent calls. It would be better to include the channelId in the cached data and compare each channel by ID. THis will also help resolve my comment above because you can selectively call the `callback` only for the channels that have changed. This will also need to handle cases where a channel has been added or removed.

package.json
```diff
@@ -55,6 +56,7 @@
     "electron-log": "4.4.8",
     "electron-window-state": "5.0.3",
     "fs-extra": "11.1.0",
+    "lodash": "^4.17.21",
```
jamaljsr, line 59: Oh wow, I thought we already had `lodash` in the project. That's why I suggested it. Thanks for adding it.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -197,9 +199,100 @@ class CLightningService implements LightningService {
     );
   }
 
+  async addListenerToNode(node: LightningNode): Promise<void> {
+    console.log('addListenerToNode CLN on port: ', node.ports.rest);
```
jamaljsr, line 203: You should use `debug` from `electron-log` instead of `console.log`. Also would be good to add a comment stating that this is a no-op because polling is used for CLN nodes.

jamaljsr (commented):
Thanks for the updates. This is getting very close to done. 

I left a bit more feedback on your latest changes. It's mostly small stuff to tidy up the code. 

The only other things to address are:
- rebase this branch on `master`
- fix unit tests and fill the remaining code coverage

Once these are done, I think this will be pretty much ready to merge.

kelvinator07:
> Thanks for the updates. This is getting very close to done.
> 
> I left a bit more feedback on your latest changes. It's mostly small stuff to tidy up the code.
> 
> The only other things to address are:
> 
> * rebase this branch on `master`
> * fix unit tests and fill the remaining code coverage
> 
> Once these are done, I think this will be pretty much ready to merge.

Thanks for the feedback. Would wrap this up by the end of this week.

src/lib/lightning/clightning/clightningService.ts
kelvinator07: Updated

src/lib/lightning/clightning/clightningService.ts
kelvinator07: Updated

src/lib/lightning/clightning/clightningService.ts
kelvinator07: Removed

src/lib/lightning/eclair/eclairService.ts
```diff
@@ -250,6 +250,40 @@ class EclairService implements LightningService {
...
   }
 
+  async addListenerToNode(node: LightningNode): Promise<void> {
+    setupListener(this.cast(node));
+  }
+
+  async removeListener(node: LightningNode): Promise<void> {
+    const listener = getListener(node);
+    listener?.close();
+  }
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/lib/lightning/lnd/lndService.ts
```diff
@@ -186,12 +186,43 @@ class LndService implements LightningService {
...
+
   private cast(node: LightningNode): LndNode {
     if (node.implementation !== 'LND')
       throw new Error(`LndService cannot be used for '${node.implementation}' nodes`);
 
     return node as LndNode;
   }
+
+  private parseData = (data: LND.ChannelEventUpdate): PLN.LightningNodeChannelEvent => {
+    if (data?.pendingOpenChannel) {
```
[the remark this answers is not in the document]
kelvinator07: Yes the data should always be defined. Updated

src/lib/lightning/lnd/lndProxyClient.ts
```diff
@@ -63,6 +63,13 @@ class LndProxyClient {
   async decodeInvoice(node: LndNode, req: LND.PayReqString): Promise<LND.PayReq> {
     return await this.ipc(ipcChannels.decodeInvoice, { node, req });
   }
+
+  async subscribeChannelEvents(
+    node: LndNode,
+    callback: (event: any) => void,
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/lib/lightning/lnd/lndService.ts
```diff
@@ -186,12 +186,43 @@ class LndService implements LightningService {
...
+      return { type: 'Pending' };
+    }
+    if (data?.activeChannel?.fundingTxidBytes) {
+      return { type: 'Open' };
+    }
+    if (data?.inactiveChannel?.fundingTxidBytes || data?.closedChannel) {
+      return { type: 'Closed' };
+    }
+    return { type: 'Unknown' };
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/utils/async.ts
```diff
@@ -49,3 +51,17 @@ export const waitFor = async (
...
   });
 };
+
+/**
+ * Creates a throttled function that only invokes `func` at most once per every `interval` milliseconds.
+ * @param func The asynchronous function to be throttled.
+ * @param interval The time window (in milliseconds) within which the function can only be called once. Default is 10 seconds.
+ * @returns Throttled version of the provided function.
+ */
+export const throttleFunction = async (
```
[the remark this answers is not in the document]
kelvinator07: Updated

src/store/models/lightning.ts
```diff
@@ -290,6 +293,84 @@ const lightningModel: LightningModel = {
...
+    nodes.lightning.forEach(
+      async node =>
+        await injections.lightningFactory.getService(node).removeListener(node),
+    );
+  }),
+  addChannelListeners: thunk(
+    async (actions, network, { injections, getStoreState, getStoreActions }) => {
+      const { nodes } = getStoreState().network.networkById(network.id);
+      for (const node of nodes.lightning) {
+        switch (node.implementation) {
```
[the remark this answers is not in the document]
kelvinator07: Updated

kelvinator07:
> Thanks for the updates. This is getting very close to done.
> 
> I left a bit more feedback on your latest changes. It's mostly small stuff to tidy up the code.
> 
> The only other things to address are:
> 
> * rebase this branch on `master`
> * fix unit tests and fill the remaining code coverage
> 
> Once these are done, I think this will be pretty much ready to merge.


@jamaljsr PR has been updated based on your last comments and is ready to be reviewed. Thanks

src/lib/lightning/eclair/eclairApi.ts
jamaljsr: You can at least reduce the duplication of the base `url` and auth header with the following in the `request` function:
```ts
  const config = setupConfig(node as EclairNode);
  const args = {
    url: `${config.url}/${path}`,
    method,
    body,
    headers: {
      ...config.headers,
      'Content-Type': 'application/x-www-form-urlencoded',
    },
  };
```

You'll just need to include `http://` in the `url` here in `setupConfig`. This already gets stripped out in `listen` by the `stripProtocol` call.
```ts
  const config = {
    url: `http://127.0.0.1:${eclairNode.ports.rest}`,
```

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Nice improvement. 👌

src/store/models/network.ts
```diff
@@ -679,9 +681,9 @@ const networkModel: NetworkModel = {
     if (networks.length === 0) {
       ipcRenderer.send('docker-shut-down');
     }
-    networks.forEach(async network => {
+    for (const network of networks) {
       await actions.stop(network.id);
-    });
+    }
```
jamaljsr, line 686: It was intentional to stop the networks in a `forEach` so that it would be faster by running in parallel. The intervel setup below will repeatedly check if all the networks are stopped before sending the IPC shut down message. So your change here can be reverted.

jamaljsr (commented):
Thanks for continuing to push this forward. I have a few more comments.

Also, it looks like the tests are failing due to the `WebSocket`. Can this be mocked?

src/store/models/network.ts
kelvinator07: Ohh okay. Changes Reverted

jamaljsr:
@kelvinator07 I have pushed a commit to your branch that adds the additional unit tests needed to get the code coverage back to 100%. There were a couple of places where code paths in the app were completely unreachable. I had to modify the implementation to eliminate those branches. This is one of the benefits of maintaining high coverage, it forces you to eliminate code that can never be executed.

I also did another round of testing for each implementation. The real time events are working great.

I am going to rebase this branch on `master` and once the CI completes, I will merge this PR. Thank you so much for completing this new feature. I'm looking forward to your next contribution 🎉

jamaljsr:
Oh, I also forgot to mention that in #879 I upgraded Polar to use the latest Core Lightning implementation and switched to using the embedded REST API which now supports WebSocket streaming 🙌. So the I am going to remove the polling and implement receiving events via sockets in the other PR.

jamaljsr (approved):

kelvinator07:
> Oh, I also forgot to mention that in #879 I upgraded Polar to use the latest Core Lightning implementation and switched to using the embedded REST API which now supports WebSocket streaming 🙌. So the I am going to remove the polling and implement receiving events via sockets in that PR.

Awesome! I'll check out the PR.

PR #841, opened 2024-03-01
feat(rename): Rename Nodes

### LINKED ISSUE #388: Feature Request: Rename nodes
It would be nice if we could rename each individual node. Sometimes, Alice, Bob and Carol... are too vague for practical usage.

Jem256:
Closes #388 

### Description
Adds a rename node feature that allows users to rename a node.

### Screenshots
Screenshot from 2024-03-07 11-06-34


Jem256:
@jamaljsr here is the code

jamaljsr:
The code currently on Github is missing the `lightning.renameNode()` function and is throwing an error, so I don't see the issue you're having.

image

Jem256:
> The code currently on Github is missing the `lightning.renameNode()` function and is throwing an error, so I don't see the issue you're having.
> 
> image

I've fixed all the issues. I kindly request for a review of my PR

Jem256:
> Thanks so much for working on this. When testing it, I found a few functional issues.
> polar-node-rename.mp4
> 
>     1. When the network is started and I enter a new name then click Save, the network is stopped, but the node isn't renamed. I have to click Save a second time to rename the node.
> 
>     2. After the rename, the node's open channels are no longer displayed.
> 
>     3. The file paths to the cert & macaroons are still showing the old name. We should rename this folder as well.
> 
>     4. I think it would be better to use a modal to have the user type the new name in so that we can access this from the right-click menu on each node.
> 
>     5. Rename is not available for the Bitcoin Core and TAP nodes. Shouldn't this be supported for all node types?

Thank so much for the review. I am going to address all these. I have a question though, about Number 4, should we remove the button and simply have the modal accessible the right click menu or should we have both?

jamaljsr:
> I have a question though, about Number 4, should we remove the button and simply have the modal accessible through the right click menu or should we have both?

To be consistent with the other modals, I think you should remove the sidebar dropdown menu and add a new "Rename Node" button between the "Restart" and "Advanced Options" buttons in the Actions tab.

Jem256:
Screenshot from 2024-03-25 14-03-57
Hey @jamaljsr do you have an idea why this is happening? when the app is restarted, the paths update to the new node name but the node also shows this error.

jamaljsr:
Hey @Jem256, for each node, there is a folder on the host machine that contains the node's data. It's located at `~/.polar/networks/<network-id>/volumes/<implementation>/<node-name>/`. It looks like Polar isn't renaming this dir when the node is renamed.

Jem256:
> ~/.polar/networks/<network-id>/volumes/<implementation>/<node-name>/

how can I access this dir through the code to update the name? I can't seem to figure that out

jamaljsr:
Hey @Jem256 please let me know when you feel this is ready for review and I'll take a look at it.

Jem256:
> Hey @Jem256 please let me know when you feel this is ready for review and I'll take a look at it.

Hey Jamal, It should be ready sometime this week. Finishing up a few things today. I'll ping you when it's ready

Jem256:
@jamaljsr please take a look and let me know what you think. The channel links are not yet working but I've tried to make a lot of other things work and would like to know your thoughts on my approach. The other thing I would like to highlight is that in case someone tries to rename the bitcoin nodes, the path to the node can't be updated because it doesn't have that property.

src/shared/types.ts
```diff
@@ -75,6 +75,7 @@ export interface BitcoinNode extends CommonNode {
 }
 
 export interface TapNode extends CommonNode {
+  lndName: string;
```
jamaljsr, line 78: This shouldn't be here `TapNode` is a generic type which `TapdNode` extends and has the `lndName` field on it. You should remove this and use that type instead.

The original thought here was that `tapd` may be one of many Taproot Asset nodes that Polar supports in the future. The `lndName` field is specific to the `tapd` implementation but may not be required for others. Think of it like how `LndNode` and `EclairNode` both extend `LightningNode`. So there may come a day where there's `TapdNode` and `SomeFutureTapNode` that both extend `TapNode`.

src/store/models/network.ts
```diff
@@ -53,6 +55,32 @@ export interface AutoMinerModel {
...
 }
 
+interface Paths {
+  tlsCert: string;
+  adminMacaroon: string;
+  invoiceMacaroon: string;
+  readonlyMacaroon: string;
+}
+
+interface Node {
```
jamaljsr, line 65: Why create a duplicate node type here when we have all of the node types in [types.ts](https://github.com/jamaljsr/polar/blob/master/src/shared/types.ts#L9). You can just accept `CommonNode` as the parameter then determine the correct node type based on the `implementation` field.

src/store/models/network.ts
```diff
@@ -943,6 +978,79 @@ const networkModel: NetworkModel = {
...
+      }
+
+      const networks = getState().networks;
+      const network = networks.find(n => n.id === node.networkId);
+      if (!network) throw new Error(l('networkByIdErr', { networkId: node.networkId }));
+
+      let updatedNode;
+      switch (node.type) {
+        case 'lightning':
+          updatedNode = network?.nodes.lightning.find(n => n.id === node.id) as Node;
```
jamaljsr, line 996: I think it makes sense to move this `switch` logic into the [utils/network](https://github.com/jamaljsr/polar/blob/master/src/utils/network.ts) file since that is responsible for manipulating the network object and setting the file paths. You could just have `renameNode(network, node, newName) which returns the updated `network` with the node's name and path changed.

src/store/models/network.ts
```diff
@@ -943,6 +978,79 @@ const networkModel: NetworkModel = {
 
     actions.setAutoMineMode({ id, mode });
   }),
+  renameNode: thunk(
+    async (actions, { node, newName }, { getState, getStoreActions, injections }) => {
+      if (!newName) throw new Error(l('renameErr', { newName }));
+
+      if (node.status === Status.Started) {
+        await actions.stop(node.networkId);
+      }
```
jamaljsr, line 987: This stops the network but it should also start it back up after performing the rename.

src/store/models/network.ts
```diff
@@ -943,6 +978,79 @@ const networkModel: NetworkModel = {
...
+
+      // Rename the docker volume data from disk
+      const volumeDir = node.implementation.toLocaleLowerCase().replace('-', '');
+      const oldPath = join(network.path, 'volumes', volumeDir, node.name);
+      const newPath = join(network.path, 'volumes', volumeDir, newName);
+
+      fs.rename(oldPath, newPath)
+        .then(() => console.log(`Renamed directory from ${oldPath} to ${newPath}`))
+        .catch((err: any) => console.error(`Error renaming directory: ${err}`));
```
jamaljsr, line 1035: The renaming of dirs should be done in the `dockerService` since that is also responsible for creating them. It can also handle saving the compose file as well.

src/store/models/designer.ts
```diff
@@ -213,6 +231,101 @@ const designerModel: DesignerModel = {
...
+      const updatedChart = {
+        ...chart,
+        nodes: {
+          ...chart.nodes,
+          [name]: {
+            ...chart.nodes[nodeId],
+            id: name,
+          },
+        },
+      };
```
jamaljsr, line 255: Nice! This is exactly how I would have done it.

src/store/models/designer.ts
```diff
@@ -213,6 +231,101 @@ const designerModel: DesignerModel = {
...
+
+      // Update the state with the modified chart
+      allCharts[activeId] = updatedChart;
+      // remove the old node with previous key
+      getStoreActions().designer.removeNode(nodeId);
+      getStoreActions().designer.updateBackendLink({
+        lnName: name,
+        backendName: backendName,
+      });
```
jamaljsr, line 264: Instead of calling `updateBackendLink`, this function should loop over all of `chart.links` and check if the `link.from.nodeId` or `link.to.nodeId` is equals the old name. If it does, then change the `nodeId` to the new name. This should fix the missing links.

src/store/models/designer.ts
```diff
@@ -213,6 +231,101 @@ const designerModel: DesignerModel = {
...
+    },
+  ),
+  renameBitcoinNode: thunk(
```
jamaljsr, line 267: I don't think it's necessary to have 3 dirrent rename node functions. Their implementations looks pretty much identical. I would try to condense these into one.

src/store/models/lightning.ts
```diff
@@ -89,7 +89,7 @@ export interface LightningModel {
 const lightningModel: LightningModel = {
   // state properties
   nodes: {},
-  // reducer actions (mutations allowed thx to immer)
+
```
jamaljsr, line 92: Did you intentionally remove this comment?

src/store/models/network.ts
```diff
@@ -943,6 +978,79 @@ const networkModel: NetworkModel = {
...
+      await actions.save();
+
+      // Loop through lightning nodes to update backend links
+      if (node.type === 'bitcoin') {
+        const lNodes = network?.nodes.lightning;
+        for (let i = 0; i < lNodes.length; i++) {
+          getStoreActions().designer.updateBackendLink({
+            lnName: lNodes[i].name,
+            backendName: newName,
+          });
```
jamaljsr, line 1047: As I mentioned in the designer file, I don't think you need to call `updateBackendLink` if you loops over the links and update the `nodeId` fields.

src/i18n/locales/de-DE.json
```diff
@@ -276,6 +276,7 @@
   "cmps.designer.lightning.LightningDetails.actions": "Aktionen",
   "cmps.designer.lightning.LightningDetails.waitingNotice": "Warten, bis {{implementation}} online ist",
   "cmps.designer.lightning.LightningDetails.connectError": "Verbindung zu {{implementation}} Knoten nicht möglich",
+  "cmps.designer.lightning.LightningDetails.menuRename": "Umbenennen",
```
jamaljsr, line 279: FYI, you only need to create new keys in the `en-US` file. The other languages are managed by the integration with Crowdin. It will generate the translations and update the other language files in a separate PR after this gets merged into `master`. So you can revert your changes to the non english files.

jamaljsr (commented):
Thanks so much for working on this feature. It has been requested many times in the past. You are off to a great start. I've left a bunch of feedback on where I see room for improvement.

One strange issue I noticed when testing was that I couldn't select the text in the input field of the modal. Instead, it was dragging the node on the canvas underneath the modal. I'm not sure how that's even possible but my hunch is it's related to where you have the Modal component being mounted. I mentioned some details about this in a comment below. 

https://github.com/jamaljsr/polar/assets/1356600/319b331e-26c6-49b1-956b-3c5408191331

src/store/models/lightning.ts
Jem256: No. that was a mistake. Let me add it back

src/store/models/network.ts
Jem256: Doing triggered this error:

> Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
> 1. You might have mismatching versions of React and the renderer (such as React DOM)
> 2. You might be breaking the Rules of Hooks
> 3. You might have more than one copy of React in the same app

This is probably because the `switch` has a `store` call in it, I am not sure though. I have maintained it in the `models/network.ts`  file for now and await your feedback

jamaljsr: The store call is identical in each `case` block. You can remove that from the `switch` and do it once after calling the network util function.

src/lib/docker/dockerService.ts
```diff
@@ -347,6 +349,24 @@ class DockerService implements DockerLibrary {
...
     }
   }
+
+  /**
+   * Renames the docker volume directory on disk when a node is renamed
+   * @param network the network containing the node
+   * @param node the node that's to be renamed
+   * @param newName the new name for the node and directory
+   */
+  async updateDirs(network: Network, node: CommonNode, newName: string) {
```
jamaljsr, line 359: Since this function is public, I wouldn't place it down here with the `private` functions. I'd put it under `removeNode` with the other functions that operate on a node. I'd also change the name to `renameNodeDir`.

src/lib/docker/dockerService.ts
```diff
@@ -347,6 +349,24 @@ class DockerService implements DockerLibrary {
...
+    const volumeDir = (node as AnyNode).implementation
+      .toLocaleLowerCase()
+      .replace('-', '');
+    const oldPath = join(network.path, 'volumes', volumeDir, node.name);
```
jamaljsr, line 363: You can use `const oldPath = nodePath(network, node.implementation, node.name);` to get this path.

src/lib/docker/dockerService.ts
```diff
@@ -347,6 +349,24 @@ class DockerService implements DockerLibrary {
...
+    const newPath = join(network.path, 'volumes', volumeDir, newName);
+
+    fs.rename(oldPath, newPath)
+      .then(() => console.log(`Renamed directory from ${oldPath} to ${newPath}`))
+      .catch((err: any) => console.error(`Error renaming directory: ${err}`));
```
jamaljsr, line 368: 1. This will not block execution while the dir is being renamed. So you may end up in a situation where the docker network is starting up while the rename is still processing. Use `await` instead of `.then` to wait for the rename to complete before proceeding.
2. We have [utils/files.ts](https://github.com/jamaljsr/polar/blob/master/src/utils/files.ts) which contains functions for file operations. It would be better to create a rename function there. I also prefer to use `fs-extra` for cleaner `async` operations.

src/store/models/designer.ts
```diff
@@ -213,6 +219,43 @@ const designerModel: DesignerModel = {
...
     });
   }),
+  renameNode: thunk(async (actions, { nodeId, name }, { getState, getStoreActions }) => {
+    const { allCharts, activeId } = getState();
+    const chart = allCharts[activeId];
+
+    // Ensure the node exists
+    const node = chart.nodes[nodeId];
+    if (!node) {
+      throw new Error(`${node} node with id ${nodeId} not found.`);
```
jamaljsr, line 229: `${node}` will always be `undefined` here so there's no reason to include it in the error message.

src/store/models/designer.ts
```diff
@@ -213,6 +219,43 @@ const designerModel: DesignerModel = {
...
+
+    // Iterate over all links in the chart to change nodeId
+    Object.values(chart.links).forEach(link => {
+      if (link.from.nodeId === nodeId) {
+        link.from.nodeId = name;
+      }
+      if (link.to.nodeId === nodeId) {
+        link.to.nodeId = name;
+      }
+    });
```
jamaljsr, line 240: Easy-peasy 👌 😁

src/i18n/locales/en-US.json
```diff
@@ -13,6 +13,7 @@
   "cmps.common.form.TapAssetSelect.collectible": "Collectible",
   "cmps.common.form.TapAssetSelect.balance": "Balance: {{balance}}",
   "cmps.common.form.TapAssetSelect.selectAsset": "Select Asset",
+  "cmps.common.form.RenameNodeInput.label": "Rename Node Input",
```
jamaljsr, line 16: This isn't being used anywhere and can be removed.

src/i18n/locales/en-US.json
```diff
@@ -37,6 +38,17 @@
   "cmps.common.RestartNode.stopConfirmTitle": "Are you sure you want to stop the {{name}} node?",
   "cmps.common.RestartNode.stopSuccess": "The node {{name}} has been stopped",
   "cmps.common.RestartNode.stopError": "Unable to stop the node",
+  "cmps.common.RenameNodeButton.title": "Rename Node",
+  "cmps.common.RenameNodeButton.btn": "Edit Node Name",
+  "cmps.common.RenameNodeButton.menu": "Rename Node",
```
jamaljsr, line 43: Change this to be just "Rename" in the context menu.

src/store/models/network.ts
```diff
@@ -943,6 +952,60 @@ const networkModel: NetworkModel = {
...
+      }
+
+      await injections.dockerService.updateDirs(network, node, newName);
+
+      actions.setNetworks([...networks]);
+      await actions.save();
+
+      await injections.dockerService.saveComposeFile(network);
+
+      await actions.start(node.networkId);
```
jamaljsr, line 1006: If the network is stopped and I rename a node, the network starts up afterwards. It should remain in the state it was originally in.

src/components/common/RenameNodeModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { CommonNode } from 'shared/types';
+import { useStoreActions, useStoreState } from 'store';
+import { useAsyncCallback } from 'react-async-hook';
+import { Network } from 'types';
+
+interface Props {
+  network: Network;
+}
+
+const RenameNodeModal: React.FC<Props> = ({ network }) => {
```
jamaljsr, line 13: image

The modal looks a bit empty. I'd suggest the following updates:
- Update title to be "Rename Node <name>" so the old name is displayed even after modifying the input value
- Add a label to the input: "New Node Name"
- If the network is started, add a message at the top stating "The network will be restarted to perform this operation". You can use `<Alert type="warning" message="..." />` to make it look nice.

src/lib/docker/dockerService.ts
```diff
@@ -347,6 +349,24 @@ class DockerService implements DockerLibrary {
...
+   */
+  async updateDirs(network: Network, node: CommonNode, newName: string) {
+    const volumeDir = (node as AnyNode).implementation
+      .toLocaleLowerCase()
+      .replace('-', '');
+    const oldPath = join(network.path, 'volumes', volumeDir, node.name);
+    const newPath = join(network.path, 'volumes', volumeDir, newName);
+
+    fs.rename(oldPath, newPath)
+      .then(() => console.log(`Renamed directory from ${oldPath} to ${newPath}`))
```
jamaljsr, line 367: You should use `info()` or `debug()` so the log message is saved in the log file.

jamaljsr (commented):
This is looking amazing! Great work 👍

I just have a bit more feedback on the latest changes.

Also, I noticed this error when trying to rename a Core Lightning node. Does it work fine for you?

image

src/components/common/RenameNodeModal.tsx
Jem256: > Update title to be "Rename Node " so the old name is displayed even after modifying the input value

I don't understand what you mean by this

jamaljsr: Oh I didnt realize Github hid the text `<name>` in my comment. I guess it thinks it's html. I was suggesting to change the title to include the current node's name. Ex: "Rename Node alice"

Jem256: ooh I get you now. Let me work on that

src/store/models/network.ts
```diff
@@ -943,6 +950,41 @@ const networkModel: NetworkModel = {
...
+      const network = networks.find(n => n.id === node.networkId);
+      if (!network) throw new Error(l('networkByIdErr', { networkId: node.networkId }));
+
+      await renameNode(network, node, newName);
+      await getStoreActions().designer.renameNode({
+        name: newName,
+        nodeId: oldNodeName,
+      });
+
+      await injections.dockerService.renameNodeDir(network, node, newName);
```
jamaljsr, line 974: Ah, i see why the new name is being passed to `dockerService.renameNodeDir()`. It's because the `await renameNode(network, node, newName);` is updating the name on the `node` first. You should be able to resolve the error I mentioned above by just moving the `dockerService` rename above the `await renameNode`.

jamaljsr (commented):
Thanks so much for the additional updates. This is looking really good. I found a few remaining issues, but this should be complete once these are fixed.

The last things remaining would be to rebase this branch on `master` and add/update the unit tests to fix the failures and get the coverage back up. Testing all of the code paths can be pretty tricky sometimes so please let me know if you need any assistance with this.

src/components/common/RenameNodeModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+        layout="vertical"
+        hideRequiredMark
+        colon={false}
+        initialValues={{ newNodeName: oldNodeName }}
+        onFinish={handleSubmit}
+      >
+        {node?.status === Status.Started ? (
+          <Alert type="warning" message={l('alert')} />
+        ) : null}
+        <Form.Item name="newNodeName" label={l('label')}>
```
jamaljsr, line 64: I noticed that we don't have any validation here, so entering weird characters like "alice : / ` asd" puts the network in a totally broken state. You cannot even delete the network after doing that.

Here's how to restrict the input to letters, numbers, `-`, and `_`. This 
```ts
        <Form.Item
          name="newNodeName"
          label={l('label')}
          rules={[
            { required: true, message: l('cmps.forms.required') },
            { pattern: /^[a-zA-Z0-9_-]+$/, message: l('invalidName') },
          ]}
        >
```

This `invalidName` string in `en-US.json` can be "Can only contain letters, numbers, underscores, and dashes.."

jamaljsr (commented):
Thanks for the updates. I've confirmed that all of the prior issues have been resolved.

I did discover one more problem which I described below.

The new code also needs unit test coverage. Are you able to work on that?

jamaljsr:
Yes, you can see coverage locally by running `yarn test:ci`. The output will show you which files have uncovered lines. Afterwards, you can also open the `coverage/lcov-report/index.html` file in a browser to view the same results in a nicer html format.

jamaljsr:
@Jem256 Thanks so much for getting this new feature 99% done. I'm planning a new release in the coming days and wanted to make sure this gets in it. So I've rebased this branch and added a commit with unit tests bring the coverage back to 100%. I also found a few edge cases that needed fixing. I am going to merge this once the CI passes. Thank you again for the contribution.

Jem256:
@jamaljsr thank you so much for this and for your patience and support to getting this feature this far.

PR #848, opened 2024-03-09
feat(Network): add description field 

### LINKED ISSUE #836: Feature Request: add a description field for each network

I've been using Polar a lot (it's the best!). I'm starting to have quite a lot of different networks and I often have trouble differentiating between them. It'd also be nice to be able to save some notes reminding me of what I was trying to accomplish.

Each network has a `name` field. But it might be nice to have a multi-line `description` field alongside for adding notes.

Figured I'd post an issue and see if anyone else would find this useful.

Nonnyjoe:
Closes #836 

### Description
Adds a description field to each network. On network creation users get to input the name and description of the network they intend to create, then while displaying a users network each card has a description component that displays the description of each network.

### Screenshots
Screenshot 2024-03-06 at 00 27 08


src/utils/tests/helpers.ts
```diff
@@ -181,12 +181,14 @@ export const testRepoState: DockerRepoState = {
 export const getNetwork = (
   networkId = 1,
   name?: string,
+  description?: string,
```
jamaljsr, line 184: I would suggest moving this to be the last parameter of this method so that you do not have to update all of the calls to this method in the many tests.

Another approach is to just remove this parameter here and only set the description in the tests that need to update it. Ex:
```
const network = getNetwork(1, 'test network', Status.Started, 2);
network.description = 'my description';
```

src/components/home/NetworkCard.tsx
```diff
@@ -28,6 +39,10 @@ const NetworkCard: React.FC<{ network: Network }> = ({ network }) => {
       extra={<StatusBadge status={network.status} />}
       onClick={handleClick}
     >
+      <Row>
+        <Styled.title>{`Description`}</Styled.title>
+        <Styled.description>{network.description}</Styled.description>
+      </Row>
```
jamaljsr, line 45: It doesn't look good for networks without a description set. I also don't like the height of the card varying based on how long the description is. I would suggest using an "info" icon with a tooltip next to the name, and only display it if a description is defined. 
image

src/types/index.ts
```diff
@@ -18,6 +18,7 @@ import { PolarPlatform } from 'utils/system';
 export interface Network {
   id: number;
   name: string;
+  description: string;
```
jamaljsr, line 21: For users with existing networks, this value will be `undefined` when Polar starts and loads the data from `networks.json`. We do not want there to ever be a mismatch between the types and the actual runtime values. This can cause some pretty hard to find bugs. 

I would suggest checking if this description is `undefined` at startup. If so, it should be set to an empty string so that the type is always correct.

src/components/network/NetworkView.tsx
```diff
@@ -172,6 +172,7 @@ const NetworkView: React.FC<RouteComponentProps<MatchParams>> = ({ match }) => {
       <Styled.PageHeader
         colors={theme.pageHeader}
         title={network.name}
+        description={'network.description'}
```
jamaljsr, line 175: This is causing a Typescript error because there is no `description` prop on the `PageHeader` component.

jamaljsr (commented):
Thanks for adding this new feature. I appreciate the contribution. I have some initial feedback.

I noticed that there is no way to modify a description after you initially create the network. I would think that this is something that users would want to do. I think a good place to do this is in the form displayed when you rename the node name. Add a field for the description here.
image

jamaljsr:
Hey @Nonnyjoe do you plan on continuing to work on this PR?

jamaljsr:
Replaced by #978

PR #849, opened 2024-03-10
Fix japanese

Hakkadaikon:
### Description

Hi, I'm a Japanese engineer.
Fixed some Japanese expressions for Polar.

- Fixed Japanese to be more natural
- There are parts where the same word was translated with different meanings, so the Japanese translation has been unified.

Please check.

### Steps to Test

1. Fixed Japanese translation.

2. Launch Polar:

```bash
$ yarn dev
```

3. Checked whether the changed parts are translated into Japanese as expected.


Hakkadaikon:
thx!🫂

PR #852, opened 2024-03-12
fix(OpenChannelModal): added default value for isPrivate checkbox

kelvinator07:
### Description

When you try to open a channel, the OpenChannelModal pops up and the checked box for `isPrivate` is unchecked - this should translate to false but it's undefined because it doesn't have a default value.

This PR fixes the default value for isPrivate to false.


jamaljsr:
Thanks for the fix. I tested this PR and it works great! 👍

PR #859, opened 2024-03-15
UI to balance LN channels

### LINKED ISSUE #831: feature request: add button to balance all existing channels in current network
Yes, the problem is: sometimes we need balanced channels to do some stuff while testing and developing locally, but balancing them manually takes precious time.

A button that we can click and have the channels balanced automatically (in the background by polar).

**Suggestion**

Here is a pseudo-code to balance the channels:

```javascript
button.onClick((event) => {
  for (channel of getChannels()) {
    amountPeer1 = channel.getAmountPeer1()
    amountPeer2 = channel.getAmountPeer2()
    if (amountPeer1 == amountPeer2)
      continue;
    if (amountPeer 1 > amountPeer2) {
      invoice = channel.generateInvoiceForPeer2({ amount: (amountPeer1-amountPeer)/2 })
      channel.payInvoiceToPeer2(invoice)
    } else {
      // same thing as above
    }
  }
})
```

I can try do this, but some guidance is welcomed

uwla:
Solves #831.

**NOTE**: work in progress, implementation not ready yet.

## Done

- Add new action in network store to autobalance channels.
- Add button component that dispatch that action

## To do

- Make it work
- Add tests

__Help needed__ here.

Trying to auto-balance the channel I got the error `Insufficient balance`, but this should not happen since the source has more than enough funds to pay the target and that can be confirmed by logging the amounts in the console.

I decided to open the PR anyway, so I can get help with the bug.


uwla:
I need some help here. After fixing the mentioned bug, I can add tests and finish the remaining work in the PR.

src/store/models/network.ts
```diff
@@ -922,6 +935,95 @@ const networkModel: NetworkModel = {
...
+        case 'c-lightning':
+          return clightningService;
+        case 'eclair':
+          return eclairService;
+        default:
+          throw new Error('unknown implementation');
+      }
+    };
+
+    const balanceChannel = async (
```
jamaljsr, line 956: This function is a bit too long to be declared inside of the action. I would recommend moving this to the [network util](https://github.com/jamaljsr/polar/blob/master/src/utils/network.ts) file instead. This is where I typically keep helper functions for this model.

You can encapsulate most of the logic in this helper function. Giving it the nodes, links, and channels, it can just return an array of objects containing the source, target, and amount to pay for only the rebalances that need to happen. Then here in the store, it just needs to call the RPCs to pay the amounts returned from the helper, if any..

src/store/models/network.ts
```diff
@@ -922,6 +935,95 @@ const networkModel: NetworkModel = {
...
+        '[AUTO BALANCE]: ',
+        'paying from ' + src.name + ' to ' + target.name + ' the amount ' + toPay,
+      );
+
+      const invoice = await getNodeLightningService(target).createInvoice(target, toPay);
+
+      await getNodeLightningService(src).payInvoice(src, invoice);
+    };
+
+    interface ChannelInfo {
```
jamaljsr, line 995: It's not really necessary to declare an invoice inside of a function. The purpose of interfaces is to be reusable throughout the codebase. Since it's going to only be used once here in this function, declare the structure inline.
```ts
const channelsInfo: {
  channel: LightningNodeChannel;
  fromNode: LightningNode;
  toNode: LightningNode;
}[] = [];
```

src/store/models/network.ts
```diff
@@ -922,6 +935,95 @@ const networkModel: NetworkModel = {
...
+      const id = channel.uniqueId;
+      const { to, from } = links[id];
+      const fromNode = id2Node[from.nodeId as string];
+      const toNode = id2Node[to.nodeId as string];
+      channelsInfo.push({ channel, fromNode, toNode });
+    }
+
+    for (const { channel, fromNode, toNode } of channelsInfo) {
+      await balanceChannel(channel, fromNode, toNode);
+    }
```
jamaljsr, line 1025: Actually, you can just merge these two loops into one and get rid of the `channelsInfo` variable altogether. I don't see any benefit of looping over two arrays when you can call `balanceChannel` in the first one.

src/store/models/network.ts
```diff
@@ -922,6 +935,95 @@ const networkModel: NetworkModel = {
...
+  }),
```
jamaljsr, line 1026: At the end of this function, you should call `syncChart` so that the channels display the updated balances.

```ts
await getStoreActions().designer.syncChart(network);
```

src/components/designer/AutoBalanceButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+interface Props {
+  network: Network;
+}
+
+const AutoBalanceButton: React.FC<Props> = ({ network }) => {
+  const { autoBalanceChannels } = useStoreActions(s => s.network);
+
+  const handleClick = async () => autoBalanceChannels({ id: network.id });
```
jamaljsr, line 20: I would suggest wrapping this call in a try/catch and displaying a notification after the rebalance completes so the user knows it succeeded or failed.

jamaljsr (changes requested):
> Trying to auto-balance the channel I got the error `Insufficient balance`, but this should not happen since the source has more than enough funds to pay the target and that can be confirmed by logging the amounts in the console.

I tested this PR and it's working fine for me. I'm not sure why you're getting that error on your end. Can you share more of the LND logs from one of the nodes?

Since it works for me, I left some feedback on the code.

src/components/network/NetworkActions.tsx
```diff
@@ -130,6 +131,7 @@ const NetworkActions: React.FC<Props> = ({
           </Button>
           <AutoMineButton network={network} />
           <SyncButton network={network} />
+          <AutoBalanceButton network={network} />
```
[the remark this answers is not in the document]
uwla: Ok. Dropdown menu is better.

src/store/models/network.ts
uwla: awesome idea

uwla:
@jamaljsr I addressed nearly all issues except the dropdown menu. I can implement tests for the added functionality, but I stil get the following error:

```
lndService.ts:160 Uncaught (in promise) Error: insufficient_balance
    at LndService.payInvoice (lndService.ts:160)
    at async Object.fn (lightning.ts:249)
    at async Promise.all (:3000/index 0)
    at async Object.fn (network.ts:972)
VM7596:2 Uncaught ReferenceError: process is not defined
    at Object.4043 (<anonymous>:2:13168)
    at r (<anonymous>:2:306599)
    at Object.8048 (<anonymous>:2:9496)
    at r (<anonymous>:2:306599)
    at Object.8641 (<anonymous>:2:1379)
    at r (<anonymous>:2:306599)
    at <anonymous>:2:315627
    at <anonymous>:2:324225
    at <anonymous>:2:324229
    at HTMLIFrameElement.e.onload (index.js:1)
```

You said it worked on your side, so I ask you to test again. I'll try to figure out why it is not working on my machine.

Actually, I'm getting some buggy behavior unrelated to the PR: sometimes I can't click on anything, which I suppose not to be a polar bug but a bug with electron due to my window manager.

uwla:
I'll be back to it in a few days.

jamaljsr:
Hey @uwla are you planning to continue to work on this PR?

I took a look at your latest changes and noticed what is causing the `insufficient_balance ` error. You have the `target` and `source` swapped in the [autoBalanceChannels](https://github.com/jamaljsr/polar/pull/859/files#diff-8aebe0ac57c3aeb06c8fb3dc7efd3a07b5e3c1f39b6c823f5b08839ce8fb1953R965-R969) function. The `target` should be the node creating the invoice and the `source` should pay it.

```ts
        promisesToAwait.push(
          createInvoice({ node: target, amount }).then(invoice =>
            payInvoice({ node: source, amount, invoice }),
          ),
        );
```

uwla:
I got some problems:

- [X] LND can pay to LND
- [X] LND can pay to CoreLightning
- [ ] LND cannot pay to Eclair neither Eclair can pay to LND:

```
Uncaught (in promise) Error: no_route
    at LndService.payInvoice (lndService.ts:160)
    at async Object.fn (lightning.ts:249)
    at async Promise.all (:3000/index 0)
    at async Object.fn (network.ts:971)
Uncaught ReferenceError: process is not defined
    at Object.4043 (<anonymous>:2:13168)
    at r (<anonymous>:2:306599)
    at Object.8048 (<anonymous>:2:9496)
    at r (<anonymous>:2:306599)
    at Object.8641 (<anonymous>:2:1379)
    at r (<anonymous>:2:306599)
    at <anonymous>:2:315627
    at <anonymous>:2:324225
    at <anonymous>:2:324229
    at HTMLIFrameElement.e.onload (index.js:1)
```

- CoreLightning cannot pay to CoreLightning:

```
Uncaught (in promise) Error: lightningd -32602: msatoshi parameter unnecessary
    at request (clightningApi.ts:38)
    at processTicksAndRejections (internal/process/task_queues.js:93)
    at async CLightningService.payInvoice (clightningService.ts:173)
    at async Object.fn (lightning.ts:249)
    at async Promise.all (:3000/index 0)
    at async Object.fn (network.ts:971)
```

After playing a bit, I found out I am also getting the same behavior when generating and paying invoices manually with the mentioned implementations: Create invoice -> Copy and Close -> Pay invoices.

The only which works pretty well is LND to LND.

jamaljsr:
Thanks for the details on the issues your facing. I think I've found solutions to your outstanding problems.

>  LND cannot pay to Eclair neither Eclair can pay to LND:

I had to dive pretty deep into the weeds to figure this one out. I was able to reproduce the behavior on my end as well. The problem essentially a configuration issue on the Eclair side. By default, it won't let you make a single payment that is more than 45% of the channel capacity. Check the [reference.conf](https://github.com/ACINQ/eclair/blob/35295af73cad4ccb870d496a89b3c724568202dc/eclair-core/src/main/resources/reference.conf#L118) for a more detailed explanation. It would be nice if Eclair logged a clearer error message. When I changed this to 100%, then Eclair began behaving correctly.

```
  --channel.max-htlc-value-in-flight-percent=100
```

Copy this config value into the default Eclair startup flags in [constants.ts](https://github.com/jamaljsr/polar/blob/master/src/utils/constants.ts#L174).

> * CoreLightning cannot pay to CoreLightning:
> 
> ```
> Uncaught (in promise) Error: lightningd -32602: msatoshi parameter unnecessary
>     at request (clightningApi.ts:38)
>     at processTicksAndRejections (internal/process/task_queues.js:93)
>     at async CLightningService.payInvoice (clightningService.ts:173)
>     at async Object.fn (lightning.ts:249)
>     at async Promise.all (:3000/index 0)
>     at async Object.fn (network.ts:971)
> ```

This error is actually saying that you should not provide the `amount` parameter to `payInvoice` since the invoice was created with an amount specified. The amount field should only be provided to CLN when the invoice contains a zero amount. The other implementations are more forgiving about this.

The [API Docs](https://docs.corelightning.org/reference/lightning-pay) state: "The pay RPC command attempts to find a route to the given destination, and send the funds it asks for. If the bolt11 does not contain an amount, amount_msat is required, otherwise if it is specified it must be null."

jamaljsr:
After testing this a bit more, I feel like immediately balancing the channels when the button is clicked can be a bit unexpected. It would be very cool if it instead displayed a modal with the details of the payments that will be made on each channel. Then when the Submit button in the modal is clicked, the channels would get balanced. What do you think about this idea?

uwla:
@jamaljsr I think it is a good idea.

So, we show a modal and for each channel:

- left side shows the opening node, with input field to choose amount
- followed by a slider
- right side shows the other node, with input field ot choose amount

After all channels, we add a button on the bottom "auto balance all channels", which will autobalance the preview values.

Finally, at the very bottom, a CONFIRM button an a CANCEL button. If user confirms, each channel will be updated in batch if needed. Otherwise the modal is hidden and its state reset.

What about it now?

uwla:
I can create the modal and add some tests for the coverage not to drop.

However, I have been facing an weird issue while developing polar locally: after launching the live Electron app, a lot of times I can't click with the mouse on the buttons, and the cursor icon does alter (it is the same as normal, instead of a click pointer). This has been the most headache thing, because I have to open the devtools and do something like `document.querySelector("myquery button").click()` to trigger the click and see the effects of my code

jamaljsr:
Hey, I really like the idea of the sliders. That would be very cool. I'm not sure about the input fields because that can really clutter up the limited space in the small modal. I think it might be easier to implement and more visually appealing to just display the node name and balance on each side with the slider in between. 

> However, I have been facing an weird issue while developing polar locally: after launching the live Electron app, a lot of times I can not click on buttons, and the cursor icon does not alter (it is the same as normal, instead of a click pointer). This has been the most headache thing, because I have to open the devtools and do something like `document.querySelector("myquery button").click()` to trigger the click and see the effects of my code

I have also noticed this happening on my machine whenever I save my changes and the UI hot reloads. I'm not sure where this is coming from, but my hunch is it's related to webpack. There is an invisible iframe being injected on the page which covers the whole screen so you can't click on anything.
image

I have been working around this using two methods depending on whether I want to lose my current place in the UI or not. Note: I'm on a Mac so the keyboard shortcuts may be different if you're on another OS.
1. Hit `CMD + SHIFT + C` then click anywhere in the UI. This will select the iframe element in the DevTools Elements tab. Then I right-click on the `<iframe>` DOM element then delete it. Now Im able to interact with the page normally.
2. Just hit `CMD + R`. This will reload the app and most times the iframe will be gone. 

I plan on investigating where this iframe is coming from and fixing it over the next few days. It's been pretty annoying since I've also been working on some updates lately.

jamaljsr:
After writing the above response, I did a quick google search and found the cause of the issue. It is the `react-error-overlay` package injecting the iframe. I just needed to force yarn to use v6.0.9 and that resolves the problem. I just pushed https://github.com/jamaljsr/polar/commit/4df612cc4907ba1ec766b8c11e1a681361423611 to `master`, so if you rebase this branch, you won't need to use the workarounds I suggested. Please let me know if this is still an issue for you.

uwla:
@jamaljsr I have sync my fork and now the bug is fixed.

I'm working on the dropdown menu and the modal.
It will be awesome when finished!

uwla:
Hey @jamaljsr, when creating a 250,000 sats, which is the default, I noticed the source starts with 246530 sats and the target 0 sats. Where are the missing sats?

uwla:
I have not finished the PR, but I pushed it anyway so you can see how it is going.

I added the dropdown and the modal.

I still need to: implement business logic, add tests, code styling refactoring.

jamaljsr:
Thanks for the update. I will check it out over the next few days.

uwla:
I'll add business logic today. The laast PR added UI, but it did not implement any logic

jamaljsr:
> Hey @jamaljsr, when creating a 250,000 sats channels, which is the default, I noticed if the implementation is LND, then the source starts with 246530 sats and the target 0 sats. Where are the missing sats?

Sorry, I forgot to answer this question. The balance is decreased by the fee for the commitment transaction and the amount allocated to the anchor output, since [Anchor](https://docs.lightning.engineering/the-lightning-network/payment-channels/understanding-sweeping#anchors) channels are created by default in LND. You can use `lncli listchannels` to get the commitment fee (`commit_fee`). For anchor outputs, the amount is always 330 sats.

For example, here's the output for one 250K sat channel I just created.
```
lnd@alice:/$ lncli listchannels
{
    "channels": [
        {
            "active": true,
            "remote_pubkey": "0347925fd31b4a2c6661fc0cc5e60fea9e549d5f2d94587048b4929aa6d8a2baa7",
            "channel_point": "03a75ce3208bf7d3cfc386d118638e5d0fe47c28caaa063220cda80c12dd6ec2:0",
            "chan_id": "131941395398656",
            "capacity": "250000",
            "local_balance": "246530",
            "remote_balance": "0",
            "commit_fee": "3140",
            "commit_weight": "772",
            "fee_per_kw": "2500",
            "unsettled_balance": "0",
            "total_satoshis_sent": "0",
            "total_satoshis_received": "0",
            "num_updates": "0",
            "pending_htlcs": [],
            "csv_delay": 144,
            "private": false,
            "initiator": true,
            "chan_status_flags": "ChanStatusDefault",
            "local_chan_reserve_sat": "2500",
            "remote_chan_reserve_sat": "2500",
            "static_remote_key": false,
            "commitment_type": "ANCHORS",
            "lifetime": "9",
            "uptime": "9",
            "close_address": "",
            "push_amount_sat": "0",
            "thaw_height": 0,
            "local_constraints": {
                "csv_delay": 144,
                "chan_reserve_sat": "2500",
                "dust_limit_sat": "354",
                "max_pending_amt_msat": "247500000",
                "min_htlc_msat": "1",
                "max_accepted_htlcs": 483
            },
            "remote_constraints": {
                "csv_delay": 144,
                "chan_reserve_sat": "2500",
                "dust_limit_sat": "354",
                "max_pending_amt_msat": "247500000",
                "min_htlc_msat": "1",
                "max_accepted_htlcs": 483
            },
            "alias_scids": [],
            "zero_conf": false,
            "zero_conf_confirmed_scid": "0",
            "peer_alias": "bob",
            "peer_scid_alias": "0",
            "memo": ""
        }
    ]
}
```

The balance breakdown is as follows:
```
    250,000   <- Channel capacity
-     3,140   <- Commitment transaction fee
-       330   <- Anchor output
-----------
    246,530   <- Local balance
```

uwla:
Thanks @jamaljsr for the updates.

My last push had some UI improvements. I will implement business logic later today, by adding a new store action. Things are working pretty well now that I can click buttons and test things with hot-reload, after fixing the react overlay bug.

uwla:
With the last push, it is WORKING!

At least with LND nodes. It stills does not work well with Clightning and Eclair due to the mentioned bugs, which we were discussing previously.

Now, what remains:

- solve the issues with clightning and eclair
- add automated tests
- minor UI improvents: nicer buttons and autoclose the modal afterwards
- refactoring (I just want a working prototype, so I may have added functions and interfaces in the wrong places)

uwla:
I'll solve the issues with Eclair and Clightning as you explained before @jamaljsr
Let's make this cool feature happen

uwla:
Some updates here.

So, I have changed the `constant.ts` to include the flags in the Eclair node:

```
      '--channel.max-htlc-value-in-flight-percent=100',
      '--channel.max-htlc-value-in-flight-msat=5000000000000', 
```

I created a new network with 1 LND, 1 Clightning and 1 Eclair.

Still got 3 types of errors, specially when trying to transact the whole amount:

```
Unhandled Rejection (Error): balance too low
EclairService.getPaymentStatus
src/lib/lightning/eclair/eclairService.ts:229

Unhandled Rejection (Error): route not found
EclairService.getPaymentStatus
src/lib/lightning/eclair/eclairService.ts:229

Unhandled Rejection (Error): lightningd 210: Ran out of routes to try after 65 attempts: see `paystatus`
request
src/lib/lightning/clightning/clightningApi.ts:38
```

The Eclair node is unable to find a "route" to a node it is directly connected? And the Clightning too.

Also, even though I have passed the flags to the default Eclair flags, it still says "balance is too low" when the remaining balance is 0.

When leaving each channel with at least around 1000 satoshis it works and the "too low" error does not show up, but sometimes the "route not found" error does happen.

The Eclair and Clightning implementation are not good compared to LND

uwla:
@jamaljsr I have made nice UI updates, check it out. The feature is working fine except in the following scenarios: trying to set low balance value for Eclair and Clightning nodes. The rest works fine.

I'll add some unit tests.

Also, since the issue with Eclair and Clightning is not caused by my code, I need your help to address that. Even in the scenario where we are unable to address the mentioned edge cases, I think this feature is worth adding.

jamaljsr:
Hey @uwla Thanks for the updates 👌

I will test out these changes in the next day or two. I'm not sure why you're having those issues which I was able to resolve on my end. But I'll try and see what happens in my testing. I'll also provide some feedback on the UX and code.

jamaljsr (commented):
Hey @uwla, I have tested this PR on my machine and was able to do the rebalance with Eclair and CLN nodes. I just waited a few minutes before attempting it to allow the graph to sync on each node.

From a functional standpoint, this feature works great. The modal looks fantastic and the channels are correctly balanced based on my selection.

The one UX thing that doesn't feel quite right is the new "Quick Actions" dropdown menu. I feel like it adds a bit of friction when I wanted to quickly mine a block or sync the chart. Also, the hover menu of the "Auto Mine" button overlays the Balance Channels button making it difficult to get to. I'd like to spend some time working on reducing the buttons in this toolbar as a follow-up. For this PR, can you just revert the NetworkActions component back to the way it was and just add an icon-only button to rebalance the channels?
image

src/store/models/modals.ts
```diff
@@ -235,6 +253,102 @@ const modalsModel: ModalsModel = {
...
+      ...state.balanceChannels,
+      ...payload,
+    };
+  }),
+  showBalanceChannels: thunk(actions => {
+    actions.setBalanceChannels({ visible: true });
+  }),
+  hideBalanceChannels: thunk(actions => {
+    actions.setBalanceChannels({ visible: false });
+  }),
```
jamaljsr, line 267: Sorry, but it looks like you moved too much of the code into this store. Only these 3 actions should live here to control whether the modal is visible and to pass along any initial state needed. 

The other actions probably live in the `lightning` store since these deal with channels and invoices.

src/store/models/modals.ts
```diff
@@ -235,6 +253,102 @@ const modalsModel: ModalsModel = {
...
+      const { getChannels } = getStoreActions().lightning;
+      const { links } = getStoreState().designer.activeChart;
+
+      const id2Node = {} as Record<string, LightningNode>;
+      const promisesToAwait = [] as Promise<unknown>[];
+      const channelsInfo = [] as ChannelInfo[];
+
+      for (const node of network.nodes.lightning) {
+        promisesToAwait.push(
+          getChannels(node).then((nodeChannels: LightningNodeChannel[]) => {
```
jamaljsr, line 280: I mentioned in the prior review to avoid `.then` and use `async/await` instead to be consistent with the rest of the codebase. I don't see a strong reason why `.then` is necessary in any of the places you've use it in this PR.

src/store/models/modals.ts
```diff
@@ -235,6 +253,102 @@ const modalsModel: ModalsModel = {
...
+          }),
+        );
+      }
+      await Promise.all(promisesToAwait);
+
+      for (const channel of channels) {
+        const { uniqueId: id, localBalance, remoteBalance } = channel;
+        if (!links[id]) continue;
+        const from = links[id].from.nodeId;
+        const to = links[id].to.nodeId as string;
```
jamaljsr, line 292: This cast is dangerous because the `to.nodeId` could be `undefined`. It would be better to `continue` in this case.

src/store/models/network.ts
```diff
@@ -179,6 +181,14 @@ export interface NetworkModel {
   setAutoMineMode: Action<NetworkModel, { id: number; mode: AutoMineMode }>;
   setMiningState: Action<NetworkModel, { id: number; mining: boolean }>;
   mineBlock: Thunk<NetworkModel, { id: number }, StoreInjections, RootModel>;
+
+  /* */
```
jamaljsr, line 185: empty comment

src/store/models/network.ts
```diff
@@ -976,6 +986,78 @@ const networkModel: NetworkModel = {
 
     actions.setAutoMineMode({ id, mode });
   }),
+  balanceChannels: thunk(
```
jamaljsr, line 989: This also belongs in the `lightning` store.

src/i18n/locales/en-US.json
```diff
@@ -21,6 +21,12 @@
   "cmps.common.AdvancedOptionsModal.cancelBtn": "Cancel",
   "cmps.common.AdvancedOptionsModal.success": "Updated advanced options for {{name}}",
   "cmps.common.AdvancedOptionsModal.error": "Failed to update options",
+  "cmps.common.BalanceChannelsButton.btn": "Balance channels",
```
jamaljsr, line 24: Capitalize "Channels"

src/components/designer/SyncButton.tsx
```diff
@@ -17,9 +17,10 @@ const Styled = {
 
 interface Props {
   network: Network;
+  children?: ReactNode;
```
jamaljsr, line 20: It looks like `children` is leftover from your prior changes and is not needed any longer.

src/components/common/BalanceChannelsButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  network: Network;
+}
+
+const BalanceChannelsButton: React.FC<Props> = ({ network }) => {
+  const { l } = usePrefixedTranslation('cmps.common.BalanceChannelsButton');
+  const { showBalanceChannels, resetBalanceChannels } = useStoreActions(s => s.modals);
+
+  return (
+    <Styled.Button
+      onClick={() => resetBalanceChannels(network).then(showBalanceChannels)}
```
jamaljsr, line 25: Drop the `.then`. If you want to reset the balances when the modal is first shown, you can wrap it in a `useAsync`, similar to the how the `OpenChannelModal` gets wallet balances.

src/components/common/BalanceChannelsButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    >
+      <SwapOutlined /> {l('btn')}
```
jamaljsr, line 27: Please remove the inline text and show the message in a tooltip, similar to the `SyncButton`.

src/components/common/BalanceChannelsModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { usePrefixedTranslation } from 'hooks';
+import { useStoreActions, useStoreState } from 'store';
+import { ChannelInfo, Network } from 'types';
+import { format } from 'utils/units';
+
+interface Props {
+  network: Network;
+}
+
+const BalanceChannelsModal: React.FC<Props> = ({ network }) => {
```
jamaljsr, line 13: I really like how simple this component is now. 👌

src/components/common/BalanceChannelsModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const CustomModalFooter = (
+    <Row gutter={10} justify="end">
+      <Col>
+        <Button onClick={() => hideBalanceChannels()}>{l('close')}</Button>
+      </Col>
+      <Col>
+        <Button onClick={() => updateBalanceOfChannels(network)}>{l('update')}</Button>
+      </Col>
+    </Row>
+  );
```
jamaljsr, line 33: Since you removed the custom font, you shouldn't need this custom footer any longer. This should make the Update button have the primary styling.

src/components/designer/BalanceChannelsButton.tsx
```diff
@@ -0,0 +1,158 @@
...
+              />
+            </div>
+          );
+        })}
+        {/* end sliders */}
+        <br />
+        <Row gutter={10}>
+          <Col span={12}>
+            <Styled.Button onClick={updateInfo}>
+              <ReloadOutlined /> Refresh
```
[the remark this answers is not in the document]
jamaljsr: I think "Reset" makes the most sense to me.

jamaljsr (commented):
Thanks for all of the updates. This is getting closer. I left more feedback on your latest changes.

I came across this UX issue. If the network doesn't have any channels, this is what the modal looks like.

image

I think it should not display the modal in this case. It should just show an error toast message stating that there are no channels to balance. Something similar to what happens when you click on Export while the network is started.

src/store/models/modals.ts
jamaljsr: I think the ideal place would be in `BalanceChannelsModal` since this data only needs to exist while the modal is on the screen. Personally, I would have implemented this slightly differently where the values you have in `channelsInfo` would be form inputs in the component. You could use hidden fields for values that do not need to be visible to the user, like the channel id and node names. Then when the form is submitted, you'll have the new balances from the sliders along with all of the other values from the hidden fields.

src/store/models/modals.ts
jamaljsr: You do not need to push promises into an array. The `for`/`of` loop will wait for the promises to resolve before continuing execution. This would work fine:
```ts
      for (const node of network.nodes.lightning) {
        const nodeChannels = await getChannels(node);
        channels.push(...nodeChannels);
        id2Node[node.name] = node;
      }
```

jamaljsr:
@uwla Apologies on the delayed response. I got tied up with other projects. I'll review this again as soon as I can this week.

jamaljsr (commented):
Hey @uwla thanks for updating this. Apologies for taking so long to review. 

I tested the functionality and it works great! I also reviewed all of the code and don't have any further suggestions there.

The only thing left to do is to get the tests passing and coverage back up to 100%. Are you able to work on this in order to get this PR merged?

jamaljsr:
Hey @uwla, I appreciate the update. I'll work on adding the tests when time permits. I will retain your existing commits in the git history.

Jem256:
> Hey @uwla, I appreciate the update. I'll work on adding the tests when time permits. I will retain your existing commits in the git history.

Hey @jamaljsr can I help with adding the tests if you've not worked on them yet?

jamaljsr:
Hey @Jem256, yes please feel free to add the tests. Thank you 🙏

Jem256:
Hey @uwla I added tests for this feature. Please check out my PR on your branch

jamaljsr:
Thanks for wrapping this up @uwla @Jem256 @Abdulkbk 🙏

I have rebased this on `master` and just added one minor UX improvement to disable the primary button after a click in the Balance Channels modal to prevent user's from triggering the process multiple times concurrently. 

This now looks good to merge. 🚀

jamaljsr:
I am not sure how that happened. I do not recall squashing any commits in my rebase. I kept all of the existing commits intact. My sincere apologies for not confirming the commit history before merging this PR. Unfortunately, there isn't much I can do now to revert this merge. I'll definitely be more mindful of the history in the future.

tehran19r (approved):

PR #878, opened 2024-04-23
feat: add custom baseport for nodes

### LINKED ISSUE #533: Feature Request: Ability to set base ports
The base port for LND REST api for the first node is usually 8081. This port seems to conflict with other applications (e.g. local development), but in particular it conflicts with XCode build PhaseScriptExecution which runs on 8081.

> PhaseScriptExecution Start\ Packager /Users/rosssavage/Library/Developer/Xcode/DerivedData/Satimoto-evtdshsaqaokluevsdbhyvsxutjv/Build/Intermediates.noindex/Satimoto.build/DebugTestnet-iphonesimulator/Satimoto.build/Script-FD10A7F022414F080027D42C.sh (in target 'Satimoto' from project 'Satimoto')
    cd /Users/rosssavage/Source/Projects/react-mobile/ios
    /bin/sh -c /Users/rosssavage/Library/Developer/Xcode/DerivedData/Satimoto-evtdshsaqaokluevsdbhyvsxutjv/Build/Intermediates.noindex/Satimoto.build/DebugTestnet-iphonesimulator/Satimoto.build/Script-FD10A7F022414F080027D42C.sh
**Connection to localhost port 8081 [tcp/sunproxyadmin] succeeded!
Port 8081 already in use, packager is either not running or not running correctly
Command PhaseScriptExecution failed with a nonzero exit code**

To be able to configure the base port settings manually either in the settings.json file or via UI

Abdulkbk:
This commit implements the feature that lets users set custom base ports for their nodes to avoid having port conflicts with other applications running on the same machine.

Closes #533 

### Description

I added a settings page with input fields to set the base ports and then save. The custom base ports persist in the settings.json file and are loaded whenever polar starts. If a node, for example, `LND` does not have a custom port set, the defaults from [constant.ts](https://github.com/jamaljsr/polar/blob/master/src/utils/constants.ts) are used

### Steps to Test

1. On the top navbar, locate the `Network Setting` tab and click on it.
2. The settings page appears, fill in the custom ports and click on save.
3. Create a network and check the ports of your nodes

### Screenshots
Screenshot from 2024-04-23 18-07-58

Screenshot from 2024-04-23 18-09-15

Node Alice
Screenshot from 2024-04-23 18-10-45

Node Bob
Screenshot from 2024-04-23 18-11-05


jamaljsr:
Hey @Abdulkbk Thanks for opening the PR. It looks like you're off to a great start.

My initial feedback just from looking at your screenshots is:
1. I think we should add some text in the Base Ports section explaining what these values are, because it isn't obvious to users what these are and why/when you should change them. I would suggest the following text.
   > Base Ports determine the starting port numbers that will be used for each node. When a network is started, Polar will use these values to detect the next open port on your machine to assign to the nodes. You usually do not need to change these unless you are running into port conflicts with other apps installed on your computer. If you change these values, try to maintain at decent size gap (roughly 100) between them to prevent them from overlapping as more nodes of each implementation are added to your networks.
2. The base port for Taproot Assets is missing
3. We should also allow changing the base ports for GRPC as well for the nodes that support it

Abdulkbk:
> Hey @Abdulkbk Thanks for opening the PR. It looks like you're off to a great start.
> 
> My initial feedback just from looking at your screenshots is:
> 1. I think we should add some text in the Base Ports section explaining what these values are, because it isn't obvious to users what these are and why/when you should change them. I would suggest the following text.
>    > Base Ports determine the starting port numbers that will be used for each node. When a network is started, Polar will use these values to detect the next open port on your machine to assign to the nodes. You usually do not need to change these unless you are running into port conflicts with other apps installed on your computer. If you change these values, try to maintain at decent size gap (roughly 100) between them to prevent them from overlapping as more nodes of each implementation are added to your networks.
> 2. The base port for Taproot Assets is missing
> 3. We should also allow changing the base ports for GRPC as well for the nodes that support it

Thanks for the feedback @jamaljsr. I'll work on the suggestions in next iteration.

Abdulkbk:
This is how it looks now @jamaljsr 

Screenshot from 2024-05-04 02-41-44

src/store/models/app.ts
```diff
@@ -79,6 +79,13 @@ const appModel: AppModel = {
       managed: [],
       custom: [],
     },
+    basePorts: {
+      LND: { grpc: null, rest: null },
+      bitcoind: { rest: null },
+      'c-lightning': { grpc: null, rest: null },
+      eclair: { rest: null },
+      tapd: { grpc: null, rest: null },
```
jamaljsr, line 87: If you set these default values to use the values from [BasePorts](https://github.com/jamaljsr/polar/blob/master/src/utils/constants.ts#L47) then we won't ever need to worry about checking for `null` or `undefined` anywhere in the app.

src/components/common/NavMenu.tsx
```diff
@@ -1,14 +1,24 @@
...
+  NETWORK_IMPORT,
+  NETWORK_NEW,
+  NODE_IMAGES,
+  NETWORK_SETTING,
+} from 'components/routing';
 
 const Styled = {
   Menu: styled.div`
-    width: 496px;
+    width: 600px;
```
jamaljsr, line 21: I suggest dropping this down to `550px` to reduce some of the white space on the right side of the menu.

jamaljsr (commented):
The UI is looking great now. I left some feedback on the code. Once those changes are made, we just need to have the unit tests updated to prevent the failures and get the coverage back up. Then this should be all set to merge.

Abdulkbk:
> The UI is looking great now. I left some feedback on the code. Once those changes are made, we just need to have the unit tests updated to prevent the failures and get the coverage back up. Then this should be all set to merge.

@jamaljsr I added a new commit addressing the feedbacks you left.

jamaljsr:
Hey @Abdulkbk I want to get this merged. Instead of going back and forth for longer, I just added an extra commit with some minor code formatting changes. I also rebased this on `master`.

Thank you for implementing this feature 🚀

PR #917, opened 2024-06-17
test(OpenChannelModal): added test case for private channel selection

kelvinator07:
The PR adds test for the scenario when`Make the channel private` checkbox is selected


jamaljsr (approved):
LGTM 🚀

PR #934, opened 2024-07-08
Core lightning channel fix

kelvinator07:

### Description

[Description of your changes go here. Please provide both casual and technical explanations.]

### Steps to Test

1. Steps
2. To
3. Test

### Screenshots

[Only if applicable]


jamaljsr:
Hey @kelvinator07, thanks so much for finding this issue. 

Do you think you'll have this PR ready in the next couple of days? I am pretty close to shipping a new release and would like to have this issue resolved before then. If you don't have the time to finish this then I can take it over to ensure it's included in the next release.

kelvinator07:
Hi @jamaljsr, The PR is now ready for review. Thanks

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -172,7 +164,7 @@ export class CLightningService implements LightningService {
   }
 
   async closeChannel(node: LightningNode, channelPoint: string): Promise<any> {
-    const body = { id: channelPoint };
+    const body = { id: channelPoint, unilateraltimeout: 1 }; // close the channel unilaterally after 1 ms
```
jamaljsr, line 167: Why add the `unilateraltimeout` parameter? This will trigger a force-close which was unexpected for me. The other nodes only do collaborative closes via the UI. If you want to force-close a channel, you need to use the CLI. I think it's best to keep that same behavior here.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -107,8 +99,8 @@ export class CLightningService implements LightningService {
         const status = ChannelStateToStatus[c.state];
         return {
           pending: status !== 'Open',
-          uniqueId: c.shortChannelId || c.channelId,
-          channelPoint: `${c.fundingTxid}:${c.fundingOutnum}`,
+          uniqueId: c.channelId.slice(-12),
+          channelPoint: c.channelId,
```
jamaljsr, line 103: This makes me a bit uncomfortable to store the channelId in the channelPoint field. I'm not sure right now how to implement this better. I'll put some thought into it and respond later.

jamaljsr (commented):
Thanks for the fix. I tested it and it works as expected. I just have some feedback on some of the implementation details.

src/lib/lightning/clightning/clightningService.ts
kelvinator07: I added the parameter because for some weird reason, Carol goes offline immediately Bob tries to close the channel. Bob keeps trying to connect with carol, and it lingers. See logs below:

```
[electron] 17:13:49.365 › Eclair API: [stream] carol {
[electron]   type: 'channel-state-changed',
[electron]   channelId: 'be815413dffd4cd5052b00df4957bc0c9059ac693c43e01d483c591f23d830e7',
[electron]   remoteNodeId: '0340abaa44e691cff8df5f35f9e97175d603097d34b956db5060014511bd3aa2ae',
[electron]   previousState: 'OFFLINE',
[electron]   currentState: 'SYNCING'
[electron] }
[electron] 17:13:49.426 › Eclair API: [stream] carol {
[electron]   type: 'channel-state-changed',
[electron]   channelId: 'be815413dffd4cd5052b00df4957bc0c9059ac693c43e01d483c591f23d830e7',
[electron]   remoteNodeId: '0340abaa44e691cff8df5f35f9e97175d603097d34b956db5060014511bd3aa2ae',
[electron]   previousState: 'SYNCING',
[electron]   currentState: 'SHUTDOWN'
[electron] }
[electron] 17:13:49.437 › Eclair API: [stream] carol {
[electron]   type: 'channel-state-changed',
[electron]   channelId: 'be815413dffd4cd5052b00df4957bc0c9059ac693c43e01d483c591f23d830e7',
[electron]   remoteNodeId: '0340abaa44e691cff8df5f35f9e97175d603097d34b956db5060014511bd3aa2ae',
[electron]   previousState: 'SHUTDOWN',
[electron]   currentState: 'OFFLINE'
[electron] }
```
I have been investigating for a while now but still couldn't find out the issue.

src/components/designer/link/Channel.spec.tsx
```diff
@@ -18,98 +18,157 @@ const bitcoindServiceMock = injections.bitcoindService as jest.Mocked<
...
+    );
+    return { ...result };
   };
 
+  const { initialState, lnd, cln } = createChannelData();
+
   describe('Channel Details', () => {
     it('should display Status', () => {
-      const { getAllByText } = renderComponent();
+      const { getAllByText } = renderComponent(initialState, lnd, cln);
```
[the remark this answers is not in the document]
kelvinator07: I have moved it into each test suite. 
The reason for the refactor is to test both scenarios when LND is the source node or CLN is the source node.
I want to be able to render a component with different source nodes.

src/lib/lightning/clightning/clightningService.ts
kelvinator07: The same is status quo for [eclair](https://github.com/jamaljsr/polar/blob/master/src/lib/lightning/eclair/eclairService.ts#L103). 
Another option will be to add a new channelId field, however, this will trickle down into a lot of files and we would need to check the node implementation in order to know which field to access for the UI, i.e channelPoint or channelId.

src/lib/lightning/clightning/clightningService.ts
jamaljsr: Ah so there appears to be a compatibility issue when closing CLN channels with an Eclair node. I tested with CLN->LND and CLN->CLN. Both of those were mutual closes. It makes sense to keep this parameter in there. Great work finding that.

Just one suggestion, the comment states `1 ms` but the [CLN Docs](https://docs.corelightning.org/reference/lightning-close) says `unilateraltimeout` is in seconds. So the comment just needs to be corrected.

jamaljsr (commented):
Thanks for those answers. This looks almost ready. I just left one remaining suggestion on the unit test changes that I think can be improved.

src/components/designer/link/Channel.spec.tsx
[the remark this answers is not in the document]
kelvinator07: Your approach is better. I've updated the test suites.

src/lib/lightning/clightning/clightningService.ts
kelvinator07: I have added a comment to address why we are force closing it, so it can be removed when the issue is solved.

jamaljsr (approved):
Thanks for the updates. This looks good to go now 👌

PR #978, opened 2024-08-14
feat(Network): Description field feature

kelvinator07:
Closes [#836](https://github.com/jamaljsr/polar/issues/836)

Replaces [#848](https://github.com/jamaljsr/polar/pull/848)

### Description

This feature adds a new optional field for the description of each network along with the network name.

### Steps to Test

1. Create a network and add the name and description (100 characters limit) in the fields provided. 
2. View the network name and description when you navigate to NetworkView.

### Home view
Screenshot 2024-08-14 at 21 06 24

### NetworkView
Screenshot 2024-08-14 at 21 07 22


jamaljsr (commented):
Thanks for implementing this. Overall everything looks great. I just have a few small suggestions for improvement.

image
- since there are now multiple fields, it would be good to have placeholders on them so it's clear what each field if for.
- I found an existing bug while testing this. If I don't enter a name, the error message is not displayed properly because the lang key doesn't exist in the `en-US.json` file. Would you mind fixing this here in this PR?

image
- I feel like the description is taking up extra space and doesn't really need to be displayed at all times. I like the tooltip on the main Networks screen. Can you use that info icon on this screen as well?

kelvinator07:
Hi @jamaljsr Thanks for your review. I'll look into the bug you mentioned and revert to the tooltip.

kelvinator07:
Hi @jamaljsr Updated the PR with the bug fix and your suggestion. Kindly review. Thanks

jamaljsr (approved):
Thanks for the updates. This looks great now!

PR #1013, opened 2024-09-26
Bug fixes

kelvinator07:
### Description

This PR fixes the following:
1. updated deprecated husky command
2. fixed error when trying to parse empty customChannelData (lnd)
3. updated returned payinvoice amount to sats (eclair)
4. removed obsolete version field from compose file


jamaljsr (approved):
Thanks for these fixes. I tested all of the changes and everything works as expected.

LGTM 🚀

PR #1056, opened 2024-11-21
Fix: Docker user warning UID is greater than SYS_UID_MAX

z017:
### Description

Currently docker entrypoints create users as system users by using the `-r` flag of `useradd`:

https://github.com/jamaljsr/polar/blob/52dc3c320b41e01e32abc6fb61bbd1606ab46ef7/docker/bitcoind/docker-entrypoint.sh#L13

The default user uid is 1000 greater than `SYS_UID_MAX` resulting in the following warning:

```
useradd warning: bitcoin's uid 1000 is greater than SYS_UID_MAX 999
```

To fix the warning:
- The `-r` flag was removed.
  - That flag also avoid home directory creation and password aging information in /etc/shadow.
- The `-M` flag was added to avoid home directory creation.
- And the password aging information is not necessary because the user has no password.

### References
- https://linux.die.net/man/8/useradd
- https://github.com/dotnet/dotnet-docker/pull/4676
  - Another project with the same issue.

### DIFF
--- a/docker/bitcoind/docker-entrypoint.sh
+++ b/docker/bitcoind/docker-entrypoint.sh
@@ -10,7 +10,7 @@ if ! id bitcoin > /dev/null 2>&1; then
 
   echo "adding user bitcoin ($USERID:$GROUPID)"
   groupadd -f -g $GROUPID bitcoin
-  useradd -r -u $USERID -g $GROUPID bitcoin
+  useradd -M -u $USERID -g $GROUPID bitcoin
   chown -R $USERID:$GROUPID /home/bitcoin
 fi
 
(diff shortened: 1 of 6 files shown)


jamaljsr:
@z017 Thanks for proposing this.

How can I reproduce this issue? I'm on and ARM Mac and do not see this warning logged when running the nodes in Polar.

z017:
I forgot to explain how I got the warning!

I'm using the docker images to build a CI test. I got the warning from the docker logs.

From a Windows Linux Subsystem running the following I can see the warning:
```sh
docker run -it --rm polarlightning/bitcoind:28.0 -regtest
```

Output:
```
...
Status: Downloaded newer image for polarlightning/bitcoind:28.0
adding user bitcoin (1000:1000)
useradd warning: bitcoin's uid 1000 is greater than SYS_UID_MAX 999
/entrypoint.sh: assuming arguments for bitcoind
Running as bitcoin user: bitcoind -regtest
2024-11-30T03:02:58Z Bitcoin Core version v28.0.0 (release build)
2024-11-30T03:02:58Z Script verification uses 7 additional threads
...
```

PD: thansk to you! Polar is an amazing project.

PR #1092, opened 2024-12-30
refactor: make bitcoin implementation flexible 

### LINKED ISSUE #302: Feature Request: btcd node
I saw that this was included in the TODO, but I would like to open an issue about it so we can discuss and keep track of it.
I would like to be able to use btcd as a backend instead of bitcoind.

As lnd makes use of Neutrino SPV for mobile and embedded devices, it's currently not possible to have a complete test environment using Polar as there's no support for the Neutrino protocol in bitcoind yet.

I could work on this if it hasn't already been started, but I cannot tell when it will be done.

Cheers
Hampus

Abdulkbk:
### Description
This PR refactors the `bitcoin` implementation.

Currently, the implementation only supports `bitcoind`, making it less flexible. This PR allows for the easy integration of other `bitcoin` implementations, such as BTCD, in the future.

**Approach:**
The PR refactors the Bitcoin implementation to utilize `BitcoinFactory`, similar to the implementation for the [Lightning](https://github.com/jamaljsr/polar/blob/master/src/lib/lightning/lightningFactory.ts) network, which allows for various Lightning implementations in Polar. This change will also improve the flexibility of the Bitcoin implementation.

### Steps to Test
Since this is a refactor PR and does not change any behavior in Polar, you can test it normally by starting Polar and creating a network that involves `bitcoind`.

Here are a few pointers to test: 
- Conduct transactions between two or more `bitcoind` nodes.
- Utilize `bitcoind` backends for some Lightning nodes and perform various operations.


jamaljsr:
Hey @Abdulkbk thanks for opening this draft PR to show what you've implemented so far. I skimmed through the diff and so far it's looking fantastic! I'm pleasantly surprised to see the test coverage still at 100%.  👌

Please let me know if you have any questions or need any assistance moving forward with this.

Abdulkbk:
Hey @jamaljsr thanks for taking the time to skim through this.

> I'm pleasantly surprised to see the test coverage still at 100%.

Yes, I try to add test to all new code as I add them👌.


> Please let me know if you have any questions or need any assistance moving forward with this.

Of course I will. I've been thinking of the best way I should add the`btcdservice` but I'm inclined towards using `REST` as I couldn't find a standard library similar to the one used for `bitcoindService`.

Abdulkbk:
Hey @jamaljsr, I am currently facing two blockers. I figured out a way for the first but I am still figuring out the second.

The first is that BTCD has wallet functionalities [decoupled](https://github.com/btcsuite/btcd/blob/09ba026580f4041bae30d732f26fa1a2036af003/docs/json_rpc_api.md?plain=1#L33) from chain functionalities, unlike `bitcoind`. However, there are workarounds. For example, I can run BTCD alongside `btcwallet`, and then use `btcwallet` for all non-CLI calls, as it supports both chain and wallet-related [RPCs](https://github.com/btcsuite/btcd/blob/09ba026580f4041bae30d732f26fa1a2036af003/docs/json_rpc_api.md?plain=1#L38).

The second blocker is that `btcwallet` [does not support regtest](https://github.com/btcsuite/btcwallet/issues/955) currently, meanwhile all the other node implementations in Polar use regtest. I think this is the real challenge.

I would love to hear your thoughts as I continue to figure a way out. Thanks.

jamaljsr:
Hi @Abdulkbk,  

> The second blocker is that `btcwallet` [does not support regtest](https://github.com/btcsuite/btcwallet/issues/955) currently, meanwhile all the other node implementations in Polar use regtest. I think this is the real challenge.
> 
> I would love to hear your thoughts as I continue to figure a way out. Thanks.

Oh wow, you are correct! `btcd` uses a different network for local testing called `simnet`. I totally forgot about this. While `lnd` does have support for simnet, the other implementations do not AFAIK. So if we move forward, we'd have to prevent adding CLN/Eclair nodes when there is no `bitcoind` present in the network. We'd also have to prevent mixing `btcd` and `bitcoind` in the same network since they won't be able to communicate with each other. This problem is going to really increase the effort required for this PR substantially due to the added compatibility complexity. 

I honestly don't have all the answers right now on how to implement this best, but I'm willing to work with you if you are still interested in pursuing this. My hunch is that there will be more challenges along the way. We'll likely need to split this up into multiple PRs so reviewing the updates won't be too difficult. Let me know if you'd like to continue with this. I'd totally understand if this feels like too large of a project to tackle at this time.

Abdulkbk:
> This problem is going to really increase the effort required for this PR substantially due to the added compatibility complexity.

I agree this will require much more effort.

> We'll likely need to split this up into multiple PRs so reviewing the updates won't be too difficult. Let me know if you'd like to continue with this. I'd totally understand if this feels like too large of a project to tackle at this time.

I would like to take this further. I will perhaps start by modifying this PR to focus on refactoring the implementation of the Bitcoin nodes to use a factory. Once I’ve completed that, I will request a review.

Abdulkbk:
Hey @jamaljsr, I think this PR is ready for reviews. Thanks

src/types/index.ts
```diff
@@ -155,6 +155,17 @@ export interface BitcoindLibrary {
   sendFunds: (node: BitcoinNode, addr: string, amount: number) => Promise<string>;
 }
 
+export interface BitcoinService {
```
jamaljsr, line 158: Can you delete the `BitcoindLibrary` interface above since this one replaces it and that one is no longer being used?

jamaljsr (changes requested):
@Abdulkbk Thank you very much for the updates to this PR. This is very well done. I wouldn't have done it any different myself. I really appreciate this update. 🙏

This looks good to go with just one small interface to cleanup. I'll be happy to merge this after that change.

src/types/index.ts
Abdulkbk: I just removed it [here](https://github.com/jamaljsr/polar/pull/1092/commits/49c1e3d412175ba3fb9bd08517cfe222c1c4f54e#diff-a9ba9cbedd19c9f66d564d2e89912890209b98f0a7ef19187877d2587300e476L147). Since we phased it out from the tests in 49c1e3d412175ba3fb9bd08517cfe222c1c4f54e commit

src/lib/bitcoin/bitcoind/bitcoindService.ts
```diff
@@ -10,12 +10,13 @@ import {
...
 } from 'utils/constants';
 
-class BitcoindService implements BitcoindLibrary {
+class BitcoindService implements BitcoinService {
   // the types in the v5 release are missing many functions so we have to cast as `any`
   // to prevent TS errors.
   createClient(node: BitcoinNode): any {
     return new BitcoinCore({
-      host: `http://127.0.0.1:${node.ports.rpc}`,
+      host: `127.0.0.1`,
```
Abdulkbk, line 18: I want also to inform you @jamaljsr  that `bitcoind` could not connect after this change until I made this modification.

Screenshot from 2025-01-27 14-20-51

Before:
```js
return new BitcoinCore({
      host: `http://127.0.0.1:${node.ports.rpc}`,
      username: bitcoinCredentials.user,
      password: bitcoinCredentials.pass,
      logger: this.log(),
      // use a long timeout due to the time it takes to mine a lot of blocks
      timeout: 5 * 60 * 1000,
    });
```

After:

```js
return new BitcoinCore({
      host: `127.0.0.1`,
      port: node.ports.rpc,
      username: bitcoinCredentials.user,
      password: bitcoinCredentials.pass,
      logger: this.log(),
      // use a long timeout due to the time it takes to mine a lot of blocks
      timeout: 5 * 60 * 1000,
    });
```

jamaljsr: Yes, the `bitcoin-core` package was updated to v5 by the bot in #1115. This was a breaking change that I fixed by modifying the constructor args. You'll need to run `yarn install` to ensure you have the latest package version which should fix the connection error without a code change.

Abdulkbk: > Yes, the `bitcoin-core` package was updated to v5 by the bot in #1115. This was a breaking change that I fixed in `master` by modifying the constructor args. You'll need to run `yarn install` to ensure you have the latest package version after the rebase, which should fix the connection error without a code change.

I just ran `yarn install` after the rebase and the connection worked without the modification. I have removed this modification now.

jamaljsr (approved):
LGTM 💪

PR #1038, opened 2024-10-29
Feat/sim ln integration

kelvinator07:
Closes #(issue number goes here)

### Description

[Description of your changes go here. Please provide both casual and technical explanations.]

### Steps to Test

1. Steps
2. To
3. Test

### Screenshots

[Only if applicable]

PR #1159, opened 2025-03-18
1/3: Add SimLn Support

### LINKED ISSUE #829: Feature Request: proposed integration of sim-ln in polar
I am opening this issue to propose the integration of [sim-ln](https://github.com/bitcoin-dev-project/sim-ln) in Polar for automated payment activity generation between running nodes.

The requirements to add sim-ln are: 
* Mount a volume containing macaroons and certs for each node
* generate a config file that outlines node access and desired activity (this is done intuitively in the UI)
* Pull the sim-ln image to run the simulation using the generated config and attached docker volume

The first step involves adding activity generation with user-defined payment flows. The simulator can also generate random activity based on the open channels.

I have started development of this feature in this [branch](https://github.com/Extheoisah/polar/tree/add-simln-activity-designer) and plan to open a draft pull request soon.

Below is the expected flow with screenshots of where development is ATM.

1. The image below shows the addition of two buttons to give users the option to choose between the `Network Designer` (default) and the `Activity Designer`
Screenshot 2024-01-17 at 21 29 59

2. When the activity designer button is clicked, you get an activity configuration designer UI (if you don't have existing activities) or a list of previously configured activities. Activities are tied to networks and cannot be used between different networks.
Screenshot 2024-01-17 at 21 29 00

3. Here's another view of already activities. The plus icon allows you to duplicate the existing configuration after saving it. You can start the simulation using the start button and it will 
     -  create a docker volume with the saved activities, 
     - create a simulation config
     - pull the sim-ln image (if not downloaded already) and run the simulation using the config
     
 
Screenshot 2024-01-17 at 21 32 48

Ideas and criticism are welcome. I hope this excites you as it does me.

Abdulkbk:
Closes #829 

Replaces #1038 

### Description
This PR is 1/3 that will integrate SimLn into Polar.

As mentioned in the previous PR implementing such:
>SimLN is a simulation tool that can be used to generate realistic payment activity on any lightning network topology. It is intentionally environment-agnostic so that it can be used across many environments - from integration tests to public signets.

>Polar provides a set of node configurations and payment activities (such as amount and frequency) to SimLN, which facilitates ongoing payments between the nodes.

### Steps to Test

1. Create a new network with 2 LND nodes. (Currently, only LND is supported; CLN is a work in progress.)
2. Start the network.
3. Open a channel between the 2 LND nodes.
4. Click on the "Simulation Designer" tab.
5. Click the "Add a new Simulation" button.
6. Select the source and destination nodes in the modal that appears.
7. Click the "Create" to add the simulation, and then Start.
8. View the Docker logs for payment activities between the nodes.

### Screenshots
Screenshot from 2025-03-18 01-24-25

Screenshot from 2025-03-18 01-25-08

Screenshot from 2025-03-18 01-25-29

I chose to divide this integration into three pull requests: 
- 1/3 adds simln support to LND.
- 2/3 introduces support for CLN.
- 3/3 implements a UI for effective reporting.

It is important to note that much of the code in this PR was taken from this [PR](https://github.com/jamaljsr/polar/pull/1038).


Abdulkbk:
Tagging @kelvinator07 and @jamaljsr for revies :)

kelvinator07:
Well Done @Abdulkbk I'll test the feature within the next couple of days and give feedback.

Abdulkbk:
cc: @kelvinator07 👀

Abdulkbk:
The project is now complete, and all LN implementations are now supported 💯

Abdulkbk:
Please review @jamaljsr when you have time. Thanks 🙏

kelvinator07:
Great work, @Abdulkbk! I really appreciate the simplicity you've brought to the feature.

Being able to run one simulation at a time and view the logs directly from the Polar UI is a fantastic user experience; awesome work, man!

I've done some manual testing, and all node permutations work well except for the Bob-to-Carol simulation. Please see the logs below for details.
```
2025-07-01 01:07:34 2025-07-01T00:07:34.624Z INFO  [simln_lib] Running the simulation forever.
2025-07-01 01:07:34 2025-07-01T00:07:34.626Z INFO  [simln_lib] Simulation is running on regtest.
2025-07-01 01:07:34 2025-07-01T00:07:34.626Z INFO  [simln_lib] Simulating 1 activity on 2 nodes.
2025-07-01 01:07:34 2025-07-01T00:07:34.626Z INFO  [simln_lib] Summary of results will be reported every 60s.
2025-07-01 01:07:34 2025-07-01T00:07:34.626Z INFO  [simln_lib] Starting activity producer for bob(032c80...2d84fe): static payment of 1000 to carol(039760...24c889) every 10s.
2025-07-01 01:07:44 2025-07-01T00:07:44.741Z ERROR [simln_lib] Error while sending payment bob(032c80...2d84fe) -> carol(039760...24c889).
2025-07-01 01:07:44 2025-07-01T00:07:44.741Z ERROR [simln_lib] Event consumer for node bob(032c80...2d84fe) exited with error: LightningError(PermanentError("\"Error calling method KeySend: RpcError { code: Some(203), message: \\\"failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)\\\", data: None }\"")).
2025-07-01 01:07:44 2025-07-01T00:07:44.741Z INFO  [simln_lib] All producers finished. Shutting down.

```

Additionally, here are a few suggestions that could improve the feature:
1. A new simulation should only be added when the network is Started.
2. A new simulation should only be added if both selected nodes have an open channel between them.
3. A simulation should only start when both selected nodes have an open channel between them.

Let me know what you think about these suggestions, if they are feasible.

jamaljsr:
@kelvinator07 thanks so much for providing the feedback. I agree with all of your suggestions.

@Abdulkbk thanks for the updates on this PR. I'll play with it this week to see if I have any additional feedback.

Abdulkbk:
> Additionally, here are a few suggestions that could improve the feature:

>A new simulation should only be added when the network is Started.
A new simulation should only be added if both selected nodes have an open channel between them. [See this](https://github.com/jamaljsr/polar/pull/1038/files#diff-8aebe0ac57c3aeb06c8fb3dc7efd3a07b5e3c1f39b6c823f5b08839ce8fb1953R1119)
A simulation should only start when both selected nodes have an open channel between them.

Thanks again for the review, @kelvinator07. All feedback has been addressed.

Abdulkbk:
review reminder @jamaljsr :)

kelvinator07:
> > Additionally, here are a few suggestions that could improve the feature:
> 
> > A new simulation should only be added when the network is Started.
> > A new simulation should only be added if both selected nodes have an open channel between them. [See this](https://github.com/jamaljsr/polar/pull/1038/files#diff-8aebe0ac57c3aeb06c8fb3dc7efd3a07b5e3c1f39b6c823f5b08839ce8fb1953R1119)
> > A simulation should only start when both selected nodes have an open channel between them.
> 
> Thanks again for the review, @kelvinator07. All feedback has been addressed.

Tested ACK! LGTM!

src/components/designer/default/SimulationDesignerTab.tsx
```diff
@@ -14,15 +30,212 @@ const Styled = {
...
 
-const SimulationDesignerTab: React.FC = () => {
+const config: {
+  [key: number]: {
+    label: string;
+    type: ButtonType;
+    danger?: boolean;
+    icon: ReactNode;
+  };
+} = {
```
jamaljsr, line 69: This is a duplicate of the var in `NetworkActions`. It would be better to reuse this instead of copying it. Maybe we can combine these two into a `StatusButton` shared component.

jamaljsr (commented):
@Abdulkbk Wow!! This is an absolutely fantastic PR. The code is great because it's very consistent with the existing codebase, the commits are structured well, the test coverage remains high. A+ job my friend 🎉 Thank you for making this review pretty easy 🙏

Of course I do have some questions and minor feedback.

1. It looks like we are only allowing 1 simulation. The plus button was disabled after adding the first one. I believe `simln` supports running multiple simulations at a time. I'm curious why you chose not to support multiple.
    Image

2. We are also restricting the activities to only two peers with a direct channel between them. Doesn't `simln` support sending between any two peers in the network? What is the reason why we have this limitation in Polar?
    Image

3. When the network is not started, the buttons to add a simulation are disabled but it's not obvious why. Can we add a message similar to what we have on the node's Connect tab? 
    Image

src/utils/constants.ts
```diff
@@ -327,6 +329,20 @@ export const dockerConfigs: Record<NodeImplementation, DockerConfig> = {
       'proofCourier',
     ],
   },
+  simln: {
+    name: 'simln',
+    imageName: 'bitcoindevproject/simln:0.2.5',
```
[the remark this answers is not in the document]
jamaljsr: > Did you build this image or is it provided by the `simln` devs? I think it makes sense to have this image under the `polarlightning` org so it uses the image we build using our `Dockerfile`. I know you don't have the permissions for that and can't change it now, but I wanted to still mention it. I'll likely add a commit to make this adjustment before merging.

src/lib/docker/composeFile.ts
```diff
@@ -194,6 +195,17 @@ class ComposeFile {
...
 
+  addSimln(networkId: number) {
+    const svc = simln(
+      dockerConfigs['simln'].name,
+      `polar-n${networkId}-simln`,
+      dockerConfigs['simln'].imageName,
+      dockerConfigs['simln'].command,
+      { ...dockerConfigs['simln'].env },
+    );
+    this.addService(svc);
```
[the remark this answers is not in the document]
Abdulkbk: Fixed

src/components/designer/default/SimulationDesignerTab.tsx
Abdulkbk: Now using a reusable shared component.

docker/simln/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+RUN apk update && apk upgrade
+
+# Copy the sim-cli binaries from the build stage to the new stage
+COPY --from=builder /sim-ln/sim-cli /usr/local/bin/
+
+#################### Polar Modification
+# Original lines:
+# COPY entrypoint.sh /entrypoint.sh
+# Copy the entrypoint script to the container
+COPY docker/entrypoint.sh /entrypoint.sh
```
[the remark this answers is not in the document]
Abdulkbk: Fixed

Abdulkbk:
I think it's safe to drop this commit d620eea1e81690ff2f38cd24bb83b9a5cb669e36 since we no longer check for channel before creating a `sim`.

I looked into why the container takes a long time to stop but couldn't find the exact cause. I think it might be related to the simln code, and I'll check further to try to fix it.

jamaljsr:
Thanks for these updates @Abdulkbk. I've tested the functionality before diving back into the code. I have some suggestions:

1. The flows to manage simulation activities feels a bit awkward: 
    1. You have to add multiple activities in the modal, then they are displayed nested under a single collapsed card in the sidebar. 
    2. Once you've added activities, the "+" icon in the sidebar remains disabled so you can't add more activities. 
    3. There's no way to edit the existing activities.
    4. Clicking on the delete menu option deletes all of the grouped activities.
    
    I would suggest changing this as follows:
    1. The modal to only add one activity at a time. 
    2. The sidebar should list each activity without any grouping/collapsing. 
    3. The "+" icon should be enabled allowing you to add more
    4. Each activity should have the menu to edit/delete them. 
    5. There should be a  "View Logs" secondary button under the status button since you'll be viewing the logs for all activities, not a single one. This button should only be visible when the simulation is started

2. It's currently possible to remove a node from the network while simulations are running. I think it the simplest solution here is when the node is being removed, if there are active running simulations that involve this node, show an error toast stating the node cannot be removedwhen it has active simulations

3. The "Amount" field in the modal should use sats instead of msats to be consistent. Just convert the value to msats when saving the simln config. I was pretty confused as to why the amounts being sent were so log.

src/types/index.ts
```diff
@@ -298,3 +303,38 @@ export interface PreInvoice {
...
+export interface Activity {
+  id: number;
+  source: LightningNode;
+  destination: LightningNode;
+  intervalSecs: number;
+  amountMsat: number;
+}
+
+export interface Simulation {
+  networkId: number;
```
jamaljsr, line 316: Do we need to store the `networkId` here? The `Simulation` object is nested inside the `Network` which already has an `id` that can be referenced.

src/components/designer/default/NetworkDesignerTab.tsx
```diff
@@ -0,0 +1,156 @@
...
+  `,
+  UpdatesButton: styled(Button)`
+    margin-top: 30px;
+  `,
+};
+
+const NetworkDesignerCard: React.FC = () => {
+  const { l } = usePrefixedTranslation(
+    'cmps.designer.default.DefaultSidebar.NetworkDesignerTab',
+  );
```
jamaljsr, line 52: - The component name should be `NetworkDesignerTab`
- The translation prefix should be `cmps.designer.default.NetworkDesignerTab`

src/components/designer/default/AddSimulationModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+      >
+        {isSameNode && <Alert type="error" message={l('sameNodesWarnMsg')} />}
+        <Row gutter={16}>
+          <Col span={12}>
+            <LightningNodeSelect
+              network={network}
+              name="source"
+              label={l('source')}
+              nodeStatus={Status.Started}
+              implementation={['LND', 'eclair', 'c-lightning']}
```
jamaljsr, line 105: Is there a technical reason why we don't allow `litd` nodes to be used for simulations?

jamaljsr (changes requested):
Thanks so much for continuing to iterate on this PR. We are getting close 💪.

I did a deep dive into the code this time around and left some comments.

Functionally this is looking pretty good, but I have some feedback:
- Deleting an activity doesn't work when I do it before starting the simulation for the first time. If I start/stop the simulation, then it does work.
- I am able to add new activities to the simulation while it's running, but this won't actually take effect until it's restarted. To avoid user confusion, I think it would be better to only allow adding new activities while the simulation is stopped.
- Large numbers are being cut off in the Amount input. I'd suggest making this field take up the full column width.
    image

jamaljsr:
Hey @Abdulkbk. Are you planning on wrapping up this PR. It feels pretty close to done. I'm going to be working on shipping a new release in the next week or two and would love to have this included. If you don't have the availability to work on this now, I can make the final updates to get this merged.

src/types/index.ts
```diff
@@ -298,3 +303,38 @@ export interface PreInvoice {
...
   nextLocalBalance: number;
 }
+
+export interface Activity {
+  id: number;
+  source: LightningNode;
+  destination: LightningNode;
+  intervalSecs: number;
+  amountMsat: number;
+}
```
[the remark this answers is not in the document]
Abdulkbk: Fixed this.

src/store/models/network.ts
```diff
@@ -1074,6 +1132,133 @@ const networkModel: NetworkModel = {
...
+            throw new Error(l('nodeNotStarted', { nodeName: source.name }));
+          if (destinationNode.status != Status.Started)
+            throw new Error(l('nodeNotStarted', { nodeName: destination.name }));
+        }
+
+        await injections.dockerService.saveComposeFile(network);
+        await injections.dockerService.startSimulation(network);
+        actions.setStatus({ id, status: Status.Started, sim: true, all: false });
+        info(`Simulation started for network '${network.name}'`);
+        await getStoreActions().app.getDockerImages();
```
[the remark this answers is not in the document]
Abdulkbk: I don't think it's necessary. I probably forgot to remove this after pulling some of the code from the previous PR.

src/components/designer/default/AddSimulationModal.tsx
Abdulkbk: Not any. I have added support for litd.

Abdulkbk:
I think the PR is ready for another round of review.

>I am able to add new activities to the simulation while it's running, but this won't actually take effect until it's restarted. To avoid user confusion, I think it would be better to only allow adding new activities while the simulation is stopped.
Large numbers are being cut off in the Amount input. I'd suggest making this field take up the full column width.

These have been addressed

> Deleting an activity doesn't work when I do it before starting the simulation for the first time. If I start/stop the simulation, then it does work.

I think this works now too but do confirm from your end.

jamaljsr (changes requested):
Thanks so much for the updates. This is looking really good. There are just two remaining issues that I'd like to address before merging this.

1. I'm still experiencing the delete activity option not working. Here's a video of what happens. This only occurs immediately after creating a new network and adding an activity. I believe this is caused by the `network.simulation` field still being undefined in `SimulationDesignerTab`.

    https://github.com/user-attachments/assets/c1929b65-0dca-4eeb-8b2e-088737c7c985

2. I'd like to resolve the long time it takes to stop the simln container. If you'd like for me to investigate this, let me know.

Abdulkbk:
> I'm still experiencing the delete activity option not working. Here's a video of what happens. This only occurs immediately after creating a new network and adding an activity. I believe this is caused by the network.simulation field still being undefined in SimulationDesignerTab.

I noticed that it was the check for `network.simulation` in `showRemoveModal` that was sometimes preventing it from working. I removed the check since we covered it in `network.ts`.

> I'd like to resolve the long time it takes to stop the simln container. If you'd like for me to investigate this, let me know.

Yes sure, in my investigation, I was unable to trace the issue to our docker-compose, and I suspected the issue might be related to the Simln codebase, which I tweaked locally, but could still not trace the root issue.

docker/simln/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+#################### Polar Modification
+# Original lines:
+# COPY entrypoint.sh /entrypoint.sh
+# Copy the entrypoint script to the container
+COPY docker-entrypoint.sh /entrypoint.sh
+####################
+
+RUN chmod +x entrypoint.sh
+
+ENTRYPOINT [ "/entrypoint.sh" ]
```
jamaljsr, line 73: I figured out how to resolve the 30sec delay when stopping the simln container. Just add this line here in the Dockerfile. 

```suggestion
RUN chmod +x entrypoint.sh

STOPSIGNAL SIGINT

ENTRYPOINT [ "/entrypoint.sh" ]
```

This tells docker to use the `SIGINT` signal when stopping a container, which typically mimics what happens when pressing `CTRL+C` in a terminal. By default docker uses the `SIGTERM` signal. I guess `sim-cli` is not trapping these signals correctly. 

After adding this one line and rebuilding the image, stopping the simulation in Polar happens very quickly.

Can you apply this change, then we'll be good to get this merged.

jamaljsr (changes requested):

docker/simln/Dockerfile
Abdulkbk: Worked like a charm! 🫡

jamaljsr:
Thanks for updating the `Dockerfile`. Apologies for the delay in approving this. Everything looks good when testing on my arm64 Mac but I also want to test on Linux, Windows, and x64 before merging this. I tried to do this today but ran into some OS issues on my other machine which I hope to resolve over the weekend. Just wanted to give you that update.

I also pushed the multi-platform docker image to Docker Hub as [polarlightning/simln](https://hub.docker.com/r/polarlightning/simln/tags). Doing this made me realize we need to update the [docker/README.md](https://github.com/jamaljsr/polar/blob/master/docker/README.md) to include instructions for building the image. Can you please add a new section underneath Lightning Terminal for simln?

jamaljsr (changes requested):
I tested this branch on Windows and ran into a problem. I still need to test on Linux.

src/lib/docker/dockerService.ts
```diff
@@ -387,6 +398,148 @@ class DockerService implements DockerLibrary {
...
+        switch (node.implementation) {
+          case 'LND':
+            const lnd = node as LndNode;
+            simNode = {
+              id: lnd.name,
+              macaroon: `/home/simln/.${lnd.paths.adminMacaroon.split('volumes/').pop()}`,
+              address: `https://host.docker.internal:${lnd.ports.grpc}`,
+              cert: `/home/simln/.${lnd.paths.tlsCert.split('volumes/').pop()}`,
+            };
+            break;
```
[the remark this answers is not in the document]
Abdulkbk: I will use the mock data in unit test. Thanks.

Abdulkbk:
@jamaljsr, are you able to conclude the testing?

jamaljsr:
Yes, I will test it again on Windows and Linux this week. Thanks for the updates.

src/lib/docker/dockerService.ts
```diff
@@ -387,6 +398,151 @@ class DockerService implements DockerLibrary {
...
+        // Split the macaroon and cert path at "volumes/" to get the relative path
+        // to the docker volume. This is necessary because the docker volumes are
+        // mounted as a different path in the container.
+        switch (node.implementation) {
+          case 'LND':
+            const lnd = node as LndNode;
+            simNode = {
+              id: lnd.name,
+              macaroon: `/home/simln/.${getPosixPath(lnd.paths.adminMacaroon)}`,
+              address: `https://host.docker.internal:${lnd.ports.grpc}`,
```
jamaljsr: Using `host.docker.internal` doesn't work on Linux. It gives me the following error when the `simln` node starts up:
```
$ docker logs -f polar-n1-simln
Error: Get info error: status: Unavailable, message: "error trying to connect: dns error: failed to lookup address information: Name does not resolve", details: [], metadata: MetadataMap { headers: {} }
```

I manually updated my `simln.json` file to use `https://polar-n1-alice:10009` and that worked on Linux. I still need to test this on the other platforms as well.

PR #1300, opened 2025-10-19
Retain custom block count input in "Manually Mine Blocks"

### LINKED ISSUE #1199: Feature Request: Retain custom block count input in "Manually Mine Blocks" on node Actions tab
When a user changes the "Manually Mine Blocks" value from the default 6 and clicks Mine, the action works correctly. However, if the user navigates away from the Actions tab and returns, the value resets to 6 instead of retaining the previously entered number.

Feature Request:

Remember the last-used block count per node in local UI state (e.g., via Redux) so users don’t have to re-enter it each time.

Why it matters:
1. Improves UX for users who prefer custom mining increments
2. Prevents accidental over-mining
3. Reduces repetitive input during testing workflows

Expected behavior:
* After entering and using a custom number, the input should persist across tab switches per node session.
* I’d love to work on this. Let me know if I can proceed, and whether there are any Redux patterns or conventions you'd like me to follow.

I’d love to work on this. Let me know if I can proceed, and whether there are any Redux patterns or conventions you'd like me to follow.

Image

Jem256:
Closes #1199

### Description

This PR adds functionality to persist the last-used custom block count in the Bitcoin → Actions tab.
Users no longer need to re-enter their preferred value for “Manually Mine Blocks” each time. Persistence is scoped per network to prevent cross-network confusion.

### Steps to Test

1. Create a new Network.
2. Navigate to the Bitcoin → Actions tab.
3. Update the default value for Manually Mine Blocks.
4. Trigger the mining action using that custom count.
5. Navigate away from the tab (or perform other UI interactions).
6. Return to the Bitcoin Actions tab. The previously entered block count should still be displayed.

### Screenshots
Screenshot 2025-10-19 211120

### DIFF
--- a/src/components/common/RenameNodeModal.spec.tsx
+++ b/src/components/common/RenameNodeModal.spec.tsx
@@ -6,9 +6,9 @@ import { initChartFromNetwork } from 'utils/chart';
 import { defaultRepoState } from 'utils/constants';
 import { createNetwork } from 'utils/network';
 import {
+  bitcoinServiceMock,
   injections,
   lightningServiceMock,
-  bitcoinServiceMock,
   litdServiceMock,
   renderWithProviders,
   tapServiceMock,
@@ -41,6 +41,7 @@ describe('RenameNodeModal', () => {
       repoState: defaultRepoState,
       managedImages: testManagedImages,
       customImages: [],
+      manualMineCount: 6,
     });
     const initialState = {
       network: {
--- a/src/components/designer/bitcoin/actions/SendOnChainModal.spec.tsx
+++ b/src/components/designer/bitcoin/actions/SendOnChainModal.spec.tsx
@@ -6,10 +6,10 @@ import { initChartFromNetwork } from 'utils/chart';
 import { defaultRepoState } from 'utils/constants';
 import { createNetwork } from 'utils/network';
 import {
+  bitcoinServiceMock,
   renderWithProviders,
   suppressConsoleErrors,
   testManagedImages,
-  bitcoinServiceMock,
 } from 'utils/tests';
 import SendOnChainModal from './SendOnChainModal';
 
@@ -31,6 +31,7 @@ describe('SendOnChainModal', () => {
       repoState: defaultRepoState,
       managedImages: testManagedImages,
       customImages: [],
+      manualMineCount: 6,
     });
 
     const initialState = {
--- a/src/components/designer/lightning/actions/CreateInvoiceModal.spec.tsx
+++ b/src/components/designer/lightning/actions/CreateInvoiceModal.spec.tsx
@@ -150,6 +150,7 @@ describe('CreateInvoiceModal', () => {
         repoState: defaultRepoState,
         managedImages: testManagedImages,
         customImages: [],
+        manualMineCount: 6,
       });
       const asset: LightningNodeChannelAsset = {
         id: 'abcd',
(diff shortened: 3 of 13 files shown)


src/components/designer/bitcoin/actions/MineBlocksInput.tsx
```diff
@@ -20,24 +20,32 @@ const Styled = {
 
 const MineBlocksInput: React.FC<{ node: BitcoinNode }> = ({ node }) => {
   const { l } = usePrefixedTranslation('cmps.designer.bitcoind.MineBlocksInput');
-  const [value, setValue] = useState(6);
+  const network = useStoreState(s => s.network.networkById(node.networkId));
   const { notify } = useStoreActions(s => s.app);
   const { mine } = useStoreActions(s => s.bitcoin);
+  const updateManualMineCount = useStoreActions(
```
Abdulkbk: can do:
 ```javascript
const {updateManualMineCount} = useStoreActions(actions => actions.network);
```

File: src/components/designer/bitcoin/actions/MineBlocksInput.tsx [hunk 2]
```diff
@@ -20,24 +20,32 @@ const Styled = {
...
+
   return (
     <Form.Item label={l('label')}>
       <InputGroup compact>
         <Styled.InputNumber
-          value={value}
+          value={network.manualMineCount}
           max={1000}
-          onChange={v => setValue(parseInt(v as any))}
+          onChange={v => handleManualMineCountChange(parseInt(v as any))}
```
Abdulkbk: Calling this function here triggers `actions.save();`, which writes to network.json every time the input changes. Maybe we should update this to `mineAsync` immediately after calling `mine` instead.

src/store/models/network.spec.ts
```diff
@@ -1023,6 +1024,52 @@ describe('Network model', () => {
...
+        "Network with the id '999' was not found.",
+      );
+    });
+
+    it('should update manual mine count and persist changes', async () => {
+      const { updateManualMineCount } = store.getActions().network;
+      const networkId = firstNetwork().id;
+
+      (injections.dockerService.saveComposeFile as jest.Mock).mockClear();
+      (injections.dockerService.saveNetworks as jest.Mock).mockClear();
```
jamaljsr: These lines are unnecessary because we configure jest with `"resetMocks": true` in `package.json. So the mocks are already cleared after every test.

jamaljsr (commented):
tACK LGTM 🚀

Just a very minor suggestion on the tests, but otherwise this looks good to go. Thanks for the updates and @Abdulkbk for the review assist. 🙏

Jem256:
> Thanks for the update, this looks good.
> 
> My only remaining suggestion on git commit cleanliness is that you shouldn't have refactoring commits in a PR. You currently have new code changes in one commit that are then reverted in a subsequent commit. This pollutes the git history making it more difficult to trace file changes in the future. It is better to remove the intermediate changes by using interactive rebase or fixup+squash commits. For example, to remove the few lines I mentioned in my prior review, You would just edit the `test(network): adding missing tests` commit instead of adding a new commit.
> 
> I am bringing this up now because this is a relatively small PR where it's easier to clean up. I'd prefer you to get comfortable with this process before you work on larger PRs. Let me know if you have any questions about this.

Thank you so much for this suggestion. I edited the commits.

jamaljsr:
Just needs a rebase then I can merge.

PR #1315, opened 2025-11-14
fix: use correct ports for the litd GRPC and REST endpoints

jcbrand:
### Description

When using a lightning-terminal node, the `Connect` tab wrongly shows the web endpoint's port for the GRPC and REST endpoints. Those endpoints have different ports.

### Steps to Test

- Add a litd node to a network.
- Click on the node and open the `Connect` tab in the right sidebar
- The GRPC and REST endpoints should have different ports from the web endpoint


jamaljsr:
Thanks for the fix. I've tested and confirmed this is a bug. 

Can you update the failing unit tests? This also needs a rebase, then I can merge it.

Issue #1212, opened 2025-05-19
Bug: eclair v0.12.0 docker image: eclair crashes when trying to open channels


ZZiigguurraatt:
eclair crashes when trying to open channels with the v0.12.0 docker image. This is with eclair <-> eclair and eclair<->lnd channels.

I am running

```
$ cat /etc/issue
Ubuntu 22.04.2 LTS \n \l
```

Here is some output from ecliar when it happens:

```
bob 2025-05-16 17:22:04,506 INFO  f.a.e.io.PeerConnection CON - connection authenticated (direction=incoming)
bob 2025-05-16 17:22:04,534 INFO  f.a.e.io.PeerConnection CON n:03c886a182d93073581d0f5b538977018dfd801abcac0d2719047fe3bca84176f7 - peer is using features=option_anchors_zero_fee_htlc_tx:optional,option_upfront_shutdown_script:optional,option_route_blinding:optional,option_data_loss_protect:mandatory,var_onion_optin:mandatory,option_static_remotekey:mandatory,payment_secret:mandatory,option_shutdown_anysegwit:optional,option_channel_type:optional,basic_mpp:optional,gossip_queries:optional (unknown=31,2023), networks=
bob 2025-05-16 17:22:04,535 INFO  f.a.e.io.PeerConnection CON n:03c886a182d93073581d0f5b538977018dfd801abcac0d2719047fe3bca84176f7 - not syncing with this peer
bob 2025-05-16 17:22:04,542 INFO  f.a.eclair.router.Router SYN n:03c886a182d93073581d0f5b538977018dfd801abcac0d2719047fe3bca84176f7 - received query_channel_range with firstBlockNum=0 numberOfBlocks=1 extendedQueryFlags_opt=TlvStream(Set(QueryFlags(1)),Set())
bob 2025-05-16 17:22:04,548 INFO  f.a.eclair.router.Router SYN n:03c886a182d93073581d0f5b538977018dfd801abcac0d2719047fe3bca84176f7 - replying with 0 items for range=(0, 1)
[...]

PR #1336, opened 2025-12-10
Tor support

### LINKED ISSUE #219: Feature request: Tor support
i want to test connecting zap iOS to `.onion` urls. would be amazing if i could just spin up an lnd node that i can connect to over tor.

Jem256:
Closes #219 

### Description

This PR adds Tor support for both Lightning Network and Bitcoin nodes, enabling users to test and use .onion services for connectivity. Provides a convenient setup for experimenting with Tor-based networks in a local or containerized environment.

**Tor Support Implementation Checklist**

- [x] Tor daemon runs per node
- [x] Tor support implemented for  Bitcoin (bitcoind), Lightning (LND, c-lightning, Eclair and Terminal )
- [x] Added a network-level toggle to enable or disable Tor for all supported nodes before starting the network
- [x] Tor configuration persists correctly across network start/stop cycles
- [x] Docker startup commands are generated dynamically based on Tor enablement
- [x] Added an onion icon to the network chart to visually identify Tor-enabled nodes
- [x] Add a per-node enable Tor option in the node submenu (trigger network restart)

### Steps to Test

Flow 1: Create network → toggle Tor ON/OFF for all nodes → start network
Flow 2: Create network → start network (TOR OFF) → modify individual node to enable Tor → restart node

### TODO
 Push new images to Docker Hub for all node versions currently supported by Polar

### Screenshots

Screenshot from 2026-01-11 18-35-45


jamaljsr:
Hey @Jem256 This is a fantastic feature. We've received many requests for Tor support over the year. Thanks you so much for tackling this.

I've done some functional testing and have a bit of feedback/questions:
1. I had to rebuild the docker images locally in order to get the nodes to start up. This is fine, but we'll need to also be sure to push new images to Docker Hub for all node versions currently supported by Polar. It would be good to add this as a TODO (for me) in the PR description since this feature won't work without doing this.
2. You currently have Tor support for Bitcoin Core, LND, and Core Lightning. Do you plan to add Terminal and Eclair, or is there some explanation for why these cannot be included?
3. What is the best way to confirm that the Tor connections are functioning properly for each node? Are there any tools/commands we can run from outside the network to verify it works? I can see the services running in the logs, but are the LN/BTC nodes actually using tor?
4. When I add multiple Bitcoin Core nodes, they are not syncing their chains with each other. When i mine blocks on one node, the block height on the others do not increase. I also don't see the onion address in the sidebar for these.
5. The Tor toggle in the NetworkActions toolbar feels too prominent for it's potential use. Since I suspect this feature will be used by a minority of users, I would prefer to place this toggle inside of the 3-dot dropdown menu instead. It's shouldn't be front & center next to the Start button.
6. I like the onion icons on each node. They look great when there are no channels, but once a node has channels, the placement of the icon feels off.
    image

Jem256:
> Hey @Jem256 This is a fantastic feature. We've received many requests for Tor support over the year. Thanks you so much for tackling this.
> 
> I've done some functional testing and have a bit of feedback/questions:
> 
> 1. I had to rebuild the docker images locally in order to get the nodes to start up. This is fine, but we'll need to also be sure to push new images to Docker Hub for all node versions currently supported by Polar. It would be good to add this as a TODO (for me) in the PR description since this feature won't work without doing this.
> 2. You currently have Tor support for Bitcoin Core, LND, and Core Lightning. Do you plan to add Terminal and Eclair, or is there some explanation for why these cannot be included?
> 3. What is the best way to confirm that the Tor connections are functioning properly for each node? Are there any tools/commands we can run from outside the network to verify it works? I can see the services running in the logs, but are the LN/BTC nodes actually using tor?
> 4. When I add multiple Bitcoin Core nodes, they are not syncing their chains with each other. When i mine blocks on one node, the block height on the others do not increase. I also don't see the onion address in the sidebar for these.
> 5. The Tor toggle in the NetworkActions toolbar feels too prominent for it's potential use. Since I suspect this feature will be used by a minority of users, I would prefer to place this toggle inside of the 3-dot dropdown menu instead. It's shouldn't be front & center next to the Start button.
> 6. I like the onion icons on each node. They look great when there are no channels, but once a node has channels, the placement of the icon feels off.
>    image

1. This has been done.
2. I have added eclair support. However, I have not found documentation on Tor regarding Terminal. Is the implementation similar to LND?
3. So far, I have been using the logs and respective node CLI to check if a Tor address is advertised. I shall add another way to test this before the next review. 
4. Since bitcoin core doesn't announce local addresses on regtest, I am still trying to figure out how we get addresses for these nodes. I think this is what is causing the issue.
5 & 6 . Will improve these

Jem256:
> I had to rebuild the docker images locally in order to get the nodes to start up. This is fine, but we'll need to also be sure to push new images to Docker Hub for all node versions currently supported by Polar. It would be good to add this as a TODO (for me) in the PR description since this feature won't work without doing this.

This was done.

> You currently have Tor support for Bitcoin Core, LND, and Core Lightning. Do you plan to add Terminal and Eclair, or is there some explanation for why these cannot be included?

All nodes are now supported. However, opening channels in Eclair is crushing. Related to #1212  This is something i am investigating. 

> When I add multiple Bitcoin Core nodes, they are not syncing their chains with each other. When i mine blocks on one node, the block height on the others do not increase. I also don't see the onion address in the sidebar for these.

Tor processes are slow and this is what affects the syncing. I have increased the delay time on `monitorstartup`  to somewhat improve this especially for peer connections. However, block propagation via tor is still slow. It eventually resolves itself but I dont know how to make it better

> The Tor toggle in the NetworkActions toolbar feels too prominent for it's potential use. Since I suspect this feature will be used by a minority of users, I would prefer to place this toggle inside of the 3-dot dropdown menu instead. It's shouldn't be front & center next to the Start button.

This was done.

> I like the onion icons on each node. They look great when there are no channels, but once a node has channels, the placement of the icon feels off.

This was improved. Channels don't throw the icon placement off anymore.

> What is the best way to confirm that the Tor connections are functioning properly for each node? Are there any tools/commands we can run from outside the network to verify it works? I can see the services running in the logs, but are the LN/BTC nodes actually using tor? 

1. `lncli --rpcserver=<grpc host> --tlscertpath=<> --macaroonpath=<> getinfo`
     Expected: find a .onion address only in the addresses field.
     `listpeers` should show .onion address for all outbound connections
     The premise of `getinfo` and looking at peers is to confirm that connections are no longer via docker bridge but   tor socks and tor is actually being used by the nodes through published addresses.
2. `torify nc -zv example.onion 9735`  (corresponding port depending on implementation)
    Expected: Connection to example.onion (127.42.42.0) 9735 port [tcp/*] succeeded! This test verifies external reachability of the ports outside the network.
    
    @jamaljsr

Jem256:
> I run a test with 2 bitcoin nodes, and they're not syncing. I run bitcoin-cli getpeerinfo, and it seems they don't see each other on the network.

This is now working. However, syncing is a bit slow because tor is slow. I'd appreciate suggestions on how to improve this

> I noticed that creating a channel when TOR is enabled is not possible, complains about insufficient funds even though the option Deposit enough sats to carol to fund the channel is selected (works when TOR is not enabled).

This is now working. 

> When a node is added to the network, if TOR is enabled globally, the new node should, by default, allow TOR.

This was implemented. 

> I noticed that when I connect an LND and a CLN as peers, LND still sees the cleartext address of the CLN lncli listpeers. Do you know if this is the correct behaviour?.

Outbound peers (`"inbound": false`) show .onion addresses with `"network": "onion"` 
`addrbind` for outbound peers is 127.0.0.1:port, meaning the connection left through the local Tor SOCKS proxy. 
Inbound peers show 127.0.0.1:port address because Tor is anonymizing the connecting peer. When a remote node connects to you via Tor, the connection arrives at your machine as a local TCP socket. 
What we don't expect to see are docker bridge IPs (eg 172.18.0.7:port)

cc @Abdulkbk

docker/lnd/docker-entrypoint.sh
```diff
@@ -14,6 +14,44 @@ if ! id lnd > /dev/null 2>&1; then
...
+DataDirectoryGroupReadable 1
+
+HiddenServiceDir /var/lib/tor/lnd-service
+HiddenServiceVersion 3
+HiddenServicePort 9735 127.0.0.1:9735
+HiddenServicePort 8080 127.0.0.1:8080
+EOF
+
+  gosu debian-tor tor &
+  TOR_PID=$!
```
Abdulkbk: are we using this anywhere?

src/shared/types.ts
```diff
@@ -25,6 +25,7 @@ export interface LightningNode extends CommonNode {
   implementation: 'LND' | 'c-lightning' | 'eclair' | 'litd';
   backendName: string;
   ports: Record<string, number | undefined>;
+  enableTor?: boolean;
```
Abdulkbk: Would it not be easier to add this to `CommonNode`, which would basically save you from type casting `(node as BitcoinNode).enableTor` since you could just do `node.enableTor`?

src/components/network/NetworkView.tsx
```diff
@@ -2,8 +2,8 @@ import React, { ReactNode, useCallback, useEffect, useState } from 'react';
 import { useAsyncCallback } from 'react-async-hook';
 import { Redirect, RouteComponentProps } from 'react-router';
 import { info } from 'electron-log';
-import styled from '@emotion/styled';
 import { InfoCircleOutlined } from '@ant-design/icons';
+import styled from '@emotion/styled';
```
Abdulkbk, line 5: Why this change?

src/lib/docker/composeFile.ts
```diff
@@ -79,6 +79,7 @@ class ComposeFile {
   }
 
   addLnd(node: LndNode, backend: CommonNode) {
+    console.log('called addLnd..');
```
Abdulkbk: nit: can remove

src/components/designer/custom/NodeInner.tsx
```diff
@@ -45,6 +46,19 @@ const CustomNodeInner: React.FC<INodeInnerDefaultProps> = ({ node }) => {
           <StatusBadge text={node.id} status={node.properties.status} />
         </span>
         <img src={node.properties.icon} style={{ width: 24, height: 24 }} alt="" />
+        {node.properties.tor && (
+          <img
+            src={torIcon}
+            style={{
```
Abdulkbk: using `Styled` is preferrable in newer code.

```javascript

const Styled = {
  Node: styled.div<{ size?: ISize; colors: ThemeColors['node']; isSelected: 
...

// The addition
  TorIcon: styled.img`
    position: absolute;
    width: 20px;
    height: 20px;
    bottom: -8px;
    right: 0;
  `,
};

// Then we can:
 {node.properties.tor && <Styled.TorIcon src={torIcon} alt="Tor" />}
```

which makes it much cleaner.

src/store/models/designer.ts
```diff
@@ -604,6 +605,27 @@ const designerModel: DesignerModel = {
...
+    (state, { payload }) => {
+      const { networkId, enabled } = payload;
+      const chart = state.allCharts[networkId];
+
+      if (chart) {
+        // Update all lightning nodes' tor property in the chart
+        Object.keys(chart.nodes).forEach(nodeName => {
+          const node = chart.nodes[nodeName];
+          // Only update lightning nodes
+          if (node.type === 'lightning') {
```
Abdulkbk: At this commit, we should've enabled it only on LND, but I guess you fixed this in a subsequent commit

src/utils/network.ts
```diff
@@ -786,23 +786,31 @@ export const getEffectiveCommand = (node: CommonNode): string => {
...
+  if (node.type === 'lightning') {
+    const lnNode = node as LightningNode;
+    return lnNode.implementation === 'LND' || lnNode.implementation === 'c-lightning';
+  }
+  if (node.type === 'bitcoin') {
+    const btcNode = node as BitcoinNode;
+    return btcNode.implementation === 'bitcoind';
+  }
+  return false;
+};
```
Abdulkbk, line 832: since only "tap" does not support tor, can we do 
```javascript
export const supportsTor = (node: CommonNode): boolean => {
    return node.type !== "tap"
}
```
??

src/store/models/bitcoin.ts
```diff
@@ -88,6 +99,34 @@ const bitcoinModel: BitcoinModel = {
       return txid;
     },
   ),
+  connectAllPeers: thunk(async (actions, network, { injections, getState }) => {
+    for (const node of network.nodes.bitcoin) {
+      try {
+        await actions.getInfo(node);
+      } catch {}
```
Abdulkbk: should we maybe log/warn if there is an error atleast?

src/utils/network.ts
```diff
@@ -690,6 +690,158 @@ export const renameNode = async (network: Network, node: AnyNode, newName: strin
...
+      return [
+        '--lnd.tor.active',
+        '--lnd.tor.v3',
+        '--lnd.tor.streamisolation',
+        '--lnd.listen=localhost',
+        '--lnd.tor.socks=127.0.0.1:9050',
+        '--lnd.tor.control=127.0.0.1:9051',
+      ];
+    case 'bitcoind':
+      return ['-proxy=127.0.0.1:9050', '-torcontrol=127.0.0.1:9051', '-bind=127.0.0.1'];
```
Abdulkbk: what does `-bind` do here?

src/utils/network.ts
```diff
@@ -690,6 +690,158 @@ export const renameNode = async (network: Network, node: AnyNode, newName: strin
...
+    lines = lines.filter(line => {
+      const trimmed = line.trim();
+      // Remove clearnet addr bindings, but keep internal Docker addr
+      return !(trimmed.startsWith('--addr=') && !trimmed.includes('statictor'));
+    });
+  }
+  let cleanCommand = lines.join('\n').trim();
+
+  if (implementation === 'bitcoind' && enableTor) {
+    cleanCommand = cleanCommand.replace('-listenonion=0', '-listenonion=1');
```
Abdulkbk: Maybe make removing `-listenonion=0` and adding `-listenonion=1` independent, else when the user modifies the default bitcoind command and removes `-listenonion=0`, this would not work.

docker/eclair/docker-entrypoint.sh
```diff
@@ -19,16 +19,56 @@ if ! id eclair > /dev/null 2>&1; then
...
+Log notice stdout
+
+SocksPort 127.0.0.1:9050
+ControlPort 127.0.0.1:9051
+CookieAuthentication 1
+CookieAuthFile /var/lib/tor/control_auth_cookie
+CookieAuthFileGroupReadable 1
+DataDirectoryGroupReadable 1
+ExitPolicy reject *:*
+
```
Abdulkbk, line 48: Is it that eclair does not require hidden service args?

For example LND has:
```bash
HiddenServiceDir /var/lib/tor/lnd-service
HiddenServiceVersion 3
HiddenServicePort 9735 127.0.0.1:9735
HiddenServicePort 8080 127.0.0.1:8080
```

docker/lnd/docker-entrypoint.sh
```diff
@@ -14,6 +14,44 @@ if ! id lnd > /dev/null 2>&1; then
   chown -R $USERID:$GROUPID /home/lnd
 fi
 
+if getent group debian-tor > /dev/null 2>&1; then
+  usermod -a -G debian-tor lnd
+fi
```
Abdulkbk: I think this usermod runs regardless of whether `ENABLE_TOR` is true or false (in all the files). Should it run only when it is true?

Abdulkbk (commented):
Looking good 👍.

I have tested with `bitcoind`, `lnd`, and `cln`. I've not tested with eclair yet. I left some feedback.

docker/eclair/docker-entrypoint.sh
Jem256: setting `eclair.tor.enabled = true` is what creates a Tor hidden service endpoint for eclair. 

[https://github.com/ACINQ/eclair/blob/master/docs/Tor.md#configure-tor-hidden-service](https://github.com/ACINQ/eclair/blob/master/docs/Tor.md#configure-tor-hidden-service)

src/utils/network.ts
Jem256: without it, you still listen on all devices and another node could establish a clearnet connection, when knowing your address. However, I realised i hadnt used it correctly. Its meant to be `-bind=127.0.0.1:8334=onion` thanks for the catch.

src/utils/network.ts
Jem256: I decided to simply remove `-listenonion=0`  since `-listenonion=1` is the default behaviour.

Abdulkbk (commented):
Nice work so far, getting there...

I tested the functionality and observed some race conditions between Tor (bootstrapping) and the startup of LN nodes. This occasionally prevents nodes from discovering or connecting to peers. For instance, I ran with three LND nodes, and sometimes peers connect successfully, while other times they do not. I suspect we need to ensure Tor has fully started and bootstrapped in docker-entry.sh before launching the LN nodes.

src/utils/network.ts
```diff
@@ -690,6 +690,58 @@ export const renameNode = async (network: Network, node: AnyNode, newName: strin
   }
 };
 
+/**
+ * Adds or removes Tor flags from an LND command
+ */
+export const applyTorFlags = (command: string, enableTor: boolean): string => {
```
Abdulkbk: I think we can do without `enableTor: boolean`. The function's name is `applyTorFlags`, and it should only be called on nodes that support Tor and have Tor enabled.

docker/bitcoind/Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BITCOIN_VERSION
 ENV PATH=/opt/bitcoin-${BITCOIN_VERSION}/bin:$PATH
 
 RUN apt-get update -y \
-  && apt-get install -y curl gosu \
+  && apt-get install -y curl gosu tor\
```
Abdulkbk: nit: add space between `tor` and `\`

docker/clightning/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG CLN_VERSION
 
 # install gosu
 RUN apt-get update -y \
-  && apt-get install -y curl gosu git \
+  && apt-get install -y curl gosu git tor\
```
Abdulkbk: nit: space between tor and `\`

src/store/models/network.ts
```diff
@@ -1323,6 +1334,53 @@ const networkModel: NetworkModel = {
...
+
+      if (wasStarted) {
+        const updatedNetwork = getState().networks.find(n => n.id === networkId);
+        if (updatedNetwork) {
+          const updatedNode = [
+            ...updatedNetwork.nodes.lightning,
+            ...updatedNetwork.nodes.bitcoin,
+          ].find(n => n.name === name);
+          if (updatedNode) {
+            await actions.toggleNode(updatedNode);
```
Abdulkbk: why not just pass the `node` supplied in this thunk?

src/utils/network.ts
```diff
@@ -790,23 +790,23 @@ export const getEffectiveCommand = (node: CommonNode): string => {
...
     }
   }
 
   return command;
 };
 
+/**
+ * Check if a node implementation supports Tor
+ */
+export const supportsTor = (node: CommonNode): boolean => {
```
Abdulkbk, line 830: Looking at this again, I think the previous implementation where we explicitly check support for all nodes has its merits, but I'm open to others' opinions.

docker/litd/Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG LITD_VERSION
 ENV PATH=/opt/litd:$PATH
 
 RUN apt-get update -y \
-  && apt-get install -y curl gosu wait-for-it \
+  && apt-get install -y curl gosu wait-for-it tor\
```
Abdulkbk: space between "tor" and "\"

src/lib/bitcoin/bitcoind/bitcoindService.ts
```diff
@@ -35,6 +35,38 @@ class BitcoindService implements BitcoinService {
...
+    let p2pHost = '';
+    const onionAddr = info.localaddresses?.find((addr: any) =>
+      addr.address.endsWith('.onion'),
+    );
+    if (onionAddr) {
+      p2pHost = `${onionAddr.address}:${onionAddr.port}`;
+    }
+    if (!p2pHost) {
+      p2pHost = `tcp://127.0.0.1:${node.ports.p2p}`;
+    }
```
Abdulkbk, line 49: can do:
```typescript
 if (onionAddr) {
      p2pHost = `${onionAddr.address}:${onionAddr.port}`;
  } else {
      p2pHost = `tcp://127.0.0.1:${node.ports.p2p}`;
  }
```

docker/eclair/docker-entrypoint.sh
```diff
@@ -19,16 +19,66 @@ if ! id eclair > /dev/null 2>&1; then
...
+  ELAPSED=0
+  until grep -q "Bootstrapped 100%" /tmp/tor.log; do
+    if [ $ELAPSED -ge $TIMEOUT ]; then
+      echo "Tor failed to bootstrap within ${TIMEOUT}s"
+      exit 1
+    fi
+    sleep 2
+    ELAPSED=$((ELAPSED + 2))
+  done
+  echo "Tor fully bootstrapped: $(cat /var/lib/tor/eclair-service/hostname)"
```
Abdulkbk: getting `cat: /var/lib/tor/eclair-service/hostname: No such file or directory`

src/store/models/bitcoin.ts
```diff
@@ -88,6 +99,45 @@ const bitcoinModel: BitcoinModel = {
...
+          if (node.enableTor && peerNode?.enableTor) {
+            const peerNodeId = getNetworkBackendId(peerNode);
+            const p2pHost = nodes[peerNodeId]?.info?.p2pHost;
+            if (p2pHost) {
+              peerAddresses.push(p2pHost);
+            }
+          } else {
+            peerAddresses.push(peerName);
+          }
+        }
```
Abdulkbk, line 135: in the `tor` branch, after checking for `const p2pHost = nodes[peerNodeId]?.info?.p2pHost;`, if `p2pHost` is still undefined, we silently skip the peer. Is that intentional or would you consider adding a fallback?.

Abdulkbk (commented):
This is really approaching the finish line quickly. I left some comments

src/store/models/network.ts
Jem256: The original `node` from the thunk argument has stale state. The re-fetch is necessary because `toggleNode`  uses node.status to decide whether to start or stop, and the original node still has `Status.Started`  after being stopped.

docker/eclair/docker-entrypoint.sh
Jem256: thanks for the catch. since eclair handles its own hidden service internally, the eclair-service directory is actually not necessary.

src/utils/network.ts
Jem256: `applyTorFlags` is also called with `enableTor: false` in `updateAdvancedOptions` to strip Tor flags from user-edited commands before saving and in `addBitcoind`, `addLnd` etc where enableTor may be false. So the parameter is needed to handle both adding and removing Tor flags. Maybe I can rename it more meaningfully. (I've renamed it to `updateTorFlags` )

src/utils/network.ts
Jem256: I understand what you mean. My reason for changing the implementation was the fact that all current implementations of lightning and bitcoin that i am aware of support Tor. So i felt doing it explicitly was regurtating this. I am also open to other opinions

src/components/network/NetworkActions.tsx
```diff
@@ -52,6 +54,65 @@ const NetworkActions: React.FC<Props> = ({
...
+          message: l(checked ? 'torEnabledAll' : 'torDisabledAll'),
+        });
+      } catch (error: any) {
+        notify({ message: l('torToggleError'), error });
+      }
+    },
+    [network.id],
+  );
+
+  const TorMenuSwitch = () => {
```
jamaljsr: This is declaring the component inside of another component. This will cause it to get recreated every re-render, which can be pretty often, and can hurt performance. It would be better to move this outside of the component and use props to pass down values from the parent.

src/lib/bitcoin/bitcoind/bitcoindService.ts
```diff
@@ -35,6 +35,37 @@ class BitcoindService implements BitcoinService {
...
+      timeoffset: info.timeoffset,
+      connections: info.connections,
+      networkactive: info.networkactive,
+      networks: info.networks,
+      relayfee: info.relayfee,
+      incrementalfee: info.incrementalfee,
+      localaddresses: info.localaddresses,
+      warnings: info.warnings,
+      p2pHost,
+    };
```
jamaljsr: Could this be written as `return { ...info, p2pHost }` to avoid the manual assignment for all of the fields.

jamaljsr (commented):
I spent more time testing this and everything is working properly. Really great work implementing this. 🙏

Just a couple minor code comments, but otherwise, this looks ready to merge.

Jem256:
@jamaljsr Thanks for the review! Comments have been addressed.

Abdulkbk:
Let's rebase this. Will take another look

Jem256:
> Let's rebase this. Will take another look

done.

PR #1339, opened 2025-12-17
Allow self-payments

jcbrand:
### Description

Adds a checkbox (under an "Advanced Options" section) which you can check to allow self payments to be made.

### Steps to Test

1. Choose "Pay Invoice"
2. Click on "Advanced Options"
3. See the "Allow self-payment" checkbox

### Screenshots

image


jamaljsr:
Thank you for this contribution. I have tested this code and one thing that stood out to me was that this will only work with LND. For CLN & Eclair, this checkbox has no effect. Do you know if those implementations support or need this same flag? 

I'm hesitant to add options to the Polar UI that isn't supported across all 3 lightning nodes. I think It will add confusion, especially if more and more options get added in the future. This is why I have intentionally kept the Polar UI pretty minimal. For advanced customizations, users would need to use the CLI which has full support for all of the available options.

jcbrand:
Thanks for you taking the time to review my PR!

We're only using LND (actually more specificically litd) because it's the only node implementation that supports Taproot Assets and all the changes I've made to Polar were related to that.

I understand the desire to keep things simple and cross-compatible between the different clients.
Ultimately it's your call whether you're wiling to have different options for different clients.

I think one can argue both ways, but I understand the argument for wanting to avoid complexity and increased maintenance.

jcbrand:
All good, thanks for a great app.

PR #1341, opened 2025-12-19
Allow tapd REST endpoint to be called from within the Docker network

jcbrand:
### Description

I've been setting up additional services inside the Polar Docker network and sometimes they need to call the nodes' API.

The certificate generated for tapd has an altname configured for the Docker host's IP address, but not for the docker container itself.

This PR fixes that.

### Steps to Test

1. Create a tapd node in Polar
2. Make an HTTP REST call from another Docker container
3. With this PR it should work, without it you should get a TLS certificate error.

@jamaljsr If you agree with this change, we should probably do the same for LND as well (and not just LITD).

### DIFF
--- a/docker/litd/docker-entrypoint.sh
+++ b/docker/litd/docker-entrypoint.sh
@@ -17,7 +17,8 @@ fi
 if [ $(echo "$1" | cut -c1) = "-" ]; then
   echo "$0: assuming arguments for litd"
 
-  set -- litd "$@"
+  CONTAINER_IP=$(hostname -i)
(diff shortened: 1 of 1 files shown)


jamaljsr:
@jcbrand  Are you still going to work on this PR?

jcbrand:
Hi @jamaljsr, I was on holiday and am only back at work this week.

I'm a bit swamped at the moment and can't look into this right now. Can you please keep this open for a few weeks then I'll try to investigate again and get back to you?

jamaljsr:
Ok np. Thanks for the update.

PR #1357, opened 2026-01-28
feat: btcd support in polar network

Abdulkbk:
Closes #302 

### Description
This PR adds btcd support to Polar.

More description is coming...

NOTE: the PR is not ready for reviews yet, it will be kep in `draft` for now.

PR #1364, opened 2026-03-03
feat: add BTCD support in polar

### LINKED ISSUE #302: Feature Request: btcd node

Abdulkbk:
Closes #302 
Replaces #1357

### Description

In this PR, we add [BTCD](https://github.com/btcsuite/btcd/), a Bitcoin full-node implementation written in Go, to Polar. Currently, only LND and litd support the btcd backend.

### Steps to Test
- Start a mix network with `bitcoind` and `btcd` nodes.

**NOTE**: Because the `btcd` and `btcwallet` images are not yet available on Docker Hub, you'll need to build them locally using the `Dockerfile` in this PR.

```bash
cd docker/btcd
docker build --build-arg BITCOIN_VERSION=0.26.0 -t polarlightning/btcd:0.26.0 .
```

```bash
cd docker/btcd/btcwallet
docker build --build-arg BTCWALLET_VERSION=0.16.13 -t polarlightning/btcwallet:0.16.13 .
```

~Notice that we're pointing to a btcd patch to build from. That's because the patch is needed before btcd can use a peer running in a separate Docker container to sync blocks. I have opened a [PR](https://github.com/btcsuite/btcd/pull/2487) addressing this.~

### Screenshots
**Network creation**:
image

**Network started**
image


src/lib/docker/composeFile.ts
```diff
@@ -202,6 +257,72 @@ class ComposeFile {
...
+
+    // // define the variable substitutions
+    const variables = {
+      rpcUser: bitcoinCredentials.user,
+      rpcPass: bitcoinCredentials.pass,
+      nodeName: backend.name,
+    };
+
+    // use the node's image
+    const image = `polarlightning/btcwallet:0.16.13`;
```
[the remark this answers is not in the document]
Abdulkbk: Btcwallet just complements btcd with wallet functionality, but the main implementation of interest is btcd.

Abdulkbk:
cc: @Jem256 for review.

src/shared/types.ts
```diff
@@ -82,6 +82,15 @@ export interface BitcoindNode extends BitcoinNode {
   };
 }
 
+export interface BtcdNode extends BitcoinNode {
+  implementation: 'btcd';
+  ports: {
+    grpc: number;
```
Jem256: `ports.grpc` is a misnomer since btcd uses JSON-RPC. This also happens in `constants.ts` and the connectTab where it's called a `GRPC Host.`

docker/btcd/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM debian:stable-slim
+
+ARG BITCOIN_VERSION
```
Jem256: Wouldn't it be better for the `BITCOIN_VERSION` argument to be `BTCD_VERSION`?

docker/btcd/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+  && mkdir /opt/btcd \
+  && mv ./btcd-linux-${SYS_ARCH}-v${BITCOIN_VERSION}/* /opt/btcd \
+  && rm *.tar.gz
+
+
+COPY docker-entrypoint.sh /entrypoint.sh
+
+RUN chmod a+x /entrypoint.sh
+
+VOLUME ["/root/.btcd"]
```
Jem256: `VOLUME ["/root/.btcd"]` doesn't match the runtime path `/home/btcd/.btcd` per the entrypoint's btcd user

src/store/models/mcp/network/createNetwork.ts
Jem256: A valid LND + btcd only network is rejected by the MCP. The `btcdVersions` needs to be read by validators

src/lib/docker/dockerService.ts
Jem256, line 449: btcd uses a nested volume path (`./volumes/btcd/{name}/btcd:/home/btcd/.btcd`, see `nodeTemplates.ts:244`), like c-lightning and litd, but `ensureDirs` only creates nodeDir and has no btcd branch, so the inner btcd dir gets auto-created by Docker as root, and the container can fail to start on Linux. `btcwallet` has the same problem: its dirs only use `ensureDir` in `startNode`, not in `start,` so starting a whole network misses them.

Suggestion: add a `btcd` branch to `ensureDirs` for the nested folder, and factor the btcwallet dir creation out so both `start()` and `startNode()` call it for every btcd node.

src/components/designer/bitcoin/ConnectTab.tsx
```diff
@@ -35,26 +35,49 @@ const ConnectTab: React.FC<Props> = ({ node }) => {
...
-  }));
+  const isBtcd = node.implementation === 'btcd';
+
+  let details: DetailValues;
+  let docsUrl: string;
+  let docsLabel: string;
+
+  if (isBtcd) {
+    // btcd-specific connection info
+    details = [
```
Jem256, line 46: The `grpcHost/walletRpcHost`  are rendered as bare `127.0.0.1:port` with no scheme, while the bitcoind branch prefixes `http://`. btcd runs with TLS on by default here (no --notls in its command), so its RPC is actually `HTTPS`. I think the display value should be `https://127.0.0.1:${port}`, not a bare host. Also worth renaming the `grpcHost` label since this is JSON-RPC, not gRPC

src/utils/network.ts
```diff
@@ -721,6 +791,7 @@ export const getMissingImages = (network: Network, pulled: string[]): string[] =
 export const getDefaultCommand = (
```
Jem256, line 803: The new optional backend param means the default command is no longer a function of (implementation, version) alone, but the `AdvancedOptionsButton.tsx` caller wasn't updated to pass it.
This creates a bug as the Advanced Options preview shows and persists bitcoind flags for a btcd-backed LND/litd node.  Opening Advanced Options on a btcd-backed node and hitting Save writes the bitcoind command onto the node and breaks it on the next start.

Jem256 (changes requested):
@Abdulkbk Thank you for this PR! I did some functional testing and made some observations below and inline comments.

1. The PR description needs updating since btcsuite/btcd#2487 was merged and v0.26.0 contains the patch. I think it would also be good to add a TODO to push the new btcd image to Docker Hub.
2. The `NewNetwork` form can't produce a btcd-only network (`bitcoindNodes` is `min={1}`), but `removeBitcoinNode` only guards `bitcoin.length === 1` so, bitcoind + btcd + CLN, then delete the `bitcoind` node, leaves a state where the next drag-and-drop crashes.

src/lib/docker/dockerService.ts
Jem256, line 1: `renameNodeDir` (line 304-333): Renaming a btcd node doesn't regenerate its TLS cert, so btcd keeps serving a cert with the old hostname; everything downstream fails the TLS handshake, and the node gets stuck at "Chain RPC is inactive." The companion `volumes/btcwallet/btcwallet-<name>` dir also never gets renamed, so you end up with a brand new empty wallet after a rename.

src/lib/bitcoin/btcd/btcdService.ts
```diff
@@ -0,0 +1,223 @@
...
+      chainwork: '0',
+      size_on_disk: 0,
+      pruneheight: 0,
+      automatic_pruning: false,
+      prune_target_size: 0,
+      softforks: result.bip9_softforks as any,
+    };
+  }
+
+  async getWalletInfo(node: BitcoinNode) {
```
Jem256, line 70: `getWalletInfo` reads balance via `getinfo`, which btcwallet proxies to btcd → balance is always 0. `sendFunds` reads that balance, thinks the wallet's empty, and `sendtoaddress` fails with "insufficient funds."

Separate issue I hit on a mixed network (one `bitcoind` backend, two `btcd` backends): the deposit itself succeeded but the UI still reported "Unable to deposit funds". `bitcoin.mine`'s `Promise.all` over every started bitcoin node means one unhealthy btcd node fails deposits for the whole network (`src/store/models/bitcoin.ts:70-81`); and `depositFunds` in `lightning.ts` always uses `nodes.bitcoin[0]` regardless of which backend the LND node is actually peered with

Jem256 (commented):
@Abdulkbk  thank you for the updates! I've left a few comments;

1. Renaming a btcd node breaks it.
2. Deposits from a btcd backend are broken end-to-end for btcd only and mixed networks.

Issue #1358, opened 2026-02-01
Bug: Route not found and usability improvement


Wgil:
Thanks for maintaining polar, I find it really helpful to understand how Lightning works behind the scenes. I wanted to give some feedback which i'm unsure if are bugs or expected. 

1. For some reason I can't figure out, after creating a channel between alice (LND) and carol (Eclair), carol cannot pay alice's invoices and I'm getting route not found.

**To Reproduce**
- Create a network with default settings.
- Create a channel between alice and carol. I also tried creating another channel from carol to alice but the issue is still there.
- Send some sats to carol if balance is empty.
- Try to create an invoice from alice and pay it from carol.

<img width="1920" height="1024" alt="Image" src="https://github.com/user-attachments/assets/1ad7c168-2c93-42fb-af0b-12d570472b0e" />

2. Another usability improvement that I found is that the app lets me pay invoices from nodes with 0 sats, Bob in this case. It should probably stop me from doing it and suggest to add some sats first.

<img width="1920" height="1036" alt="Image" src="https://github.com/user-attachments/assets/ec8bfee8-8601-4f2e-a7e8-b0a74f2f280e" />

**Desktop (please complete the following information):**
- OS: [e.g. OSX Mojave or Windows 10 v1903]
- Polar Version: v4.0.0
- Docker Version: v28.5.1
- Docker Compose Version: Compose v2.40.3-desktop.1

I can work on these if helpful.
Thanks

PR #1367, opened 2026-03-27
feat: prevent lightning invoice payments when no funds exist

Leeyah-123:
## Prevent paying invoices from nodes that have neither wallet funds nor wallet liquidity.

re #1358

### Description

This PR addresses a usability issue where the application previously allowed users to attempt paying Lightning invoices from nodes with 0 sats and no outbound liquidity, resulting in a "Sent NaN sats from ..." toast message. 

**Changes included:**
- **Pre-Payment Validation:** Added logic in the Lightning model (`lightning.ts`) to verify if a node has either a confirmed wallet balance or available outbound channel liquidity before attempting to pay an invoice.
- **Improved UI Feedback:** Updated the **Pay Invoice Modal** (`PayInvoiceModal.tsx`) to disable payment execution and display a helpful warning alert when the selected node lacks funds, guiding the user to either fund the node or open a channel with outbound liquidity first.
- **Helper Utilities:** Extracted validation logic for inbound/outbound liquidity into cleanly separated utility functions (`utils.ts`).
- **Unit Tests:** Added comprehensive test coverage in `lightning.spec.ts` to ensure the payment rejection works securely when liquidity is missing, and processes successfully when available.

### Steps to Test

1. Create a network with default settings.
2. Create an invoice with a node, and try paying with another, unfunded node.

### Screenshots

https://github.com/user-attachments/assets/8040eb1f-2ae4-47ce-940c-61f2eb4eb34d

### DIFF
--- a/src/components/designer/lightning/actions/PayInvoiceModal.tsx
+++ b/src/components/designer/lightning/actions/PayInvoiceModal.tsx
@@ -1,18 +1,19 @@
+import styled from '@emotion/styled';
+import { Alert, Form, Input, Modal, Select } from 'antd';
+import { Loader } from 'components/common';
+import AssetAmount from 'components/common/AssetAmount';
+import LightningNodeSelect from 'components/common/form/LightningNodeSelect';
+import { usePrefixedTranslation } from 'hooks';
+import { LightningNodeChannelAsset } from 'lib/lightning/types';
+
 import React, { useMemo } from 'react';
 import { useAsync, useAsyncCallback } from 'react-async-hook';
-import styled from '@emotion/styled';
-import { Form, Input, Modal, Select } from 'antd';
-import { usePrefixedTranslation } from 'hooks';
 import { LitdNode } from 'shared/types';
-import { LightningNodeChannelAsset } from 'lib/lightning/types';
 import { useStoreActions, useStoreState } from 'store';
 import { Network } from 'types';
 import { mapToTapd } from 'utils/network';
 import { ellipseInner } from 'utils/strings';
 import { format } from 'utils/units';
-import { Loader } from 'components/common';
-import AssetAmount from 'components/common/AssetAmount';
-import LightningNodeSelect from 'components/common/form/LightningNodeSelect';
 
 const Styled = {
   AssetOption: styled.div`
@@ -38,6 +39,7 @@ interface Props {
 const PayInvoiceModal: React.FC<Props> = ({ network }) => {
   const { l } = usePrefixedTranslation('cmps.designer.lightning.actions.PayInvoiceModal');
   const { visible, nodeName } = useStoreState(s => s.modals.payInvoice);
+  const { nodes } = useStoreState(s => s.lightning);
   const { getAssetsInChannels } = useStoreState(s => s.lit);
   const { formatAssetAmount } = useStoreState(s => s.tap);
   const { hidePayInvoice } = useStoreActions(s => s.modals);
@@ -48,14 +50,29 @@ const PayInvoiceModal: React.FC<Props> = ({ network }) => {
 
   const [form] = Form.useForm();
   const assetId = Form.useWatch<string>('assetId', form) || 'sats';
-  const selectedName = Form.useWatch<string>('node', form) || '';
+  const selectedName = Form.useWatch<string>('node', form) || nodeName || '';
+
+  const selectedNode = useMemo(
+    () => network.nodes.lightning.find(n => n.name === selectedName),
+    [network.nodes.lightning, selectedName],
+  );
 
   const getAssetsAsync = useAsync(async () => {
     if (!visible) return;
+
+    if (selectedNode) {
+      await getInfo(selectedNode).catch(() => undefined);
+      // Fetch every node's channels so the selected node's full
+      // outbound liquidity, including peer-opened channels, can be totaled.
+      await Promise.all(
+        network.nodes.lightning.map(node => getChannels(node).catch(() => undefined)),
+      );
+    }
+
+    // Fetch litd-specific asset data
     const litNodes = network.nodes.lightning.filter(n => n.implementation === 'litd');
     for (const node of litNodes) {
-      await getInfo(node);
-      await getChannels(node);
+      if (node.name !== selectedNode?.name) await getInfo(node).catch(() => undefined);
       await getAssetRoots(mapToTapd(node));
     }
   }, [network.nodes, visible]);
@@ -63,6 +80,31 @@ const PayInvoiceModal: React.FC<Props> = ({ network }) => {
   const assets = useMemo(() => {
     return getAssetsInChannels({ nodeName: selectedName }).map(a => a.asset);
   }, [getAssetsInChannels, selectedName]);
+  const outboundLiquidity = useMemo(() => {
+    // The selected node's spendable balance in a channel is `localBalance` when
+    // it opened the channel, or `remoteBalance` when a peer opened it (in which
+    // case the channel only appears in the peer's channel list).
+    const selfPubkey = nodes[selectedName]?.info?.pubkey;
+    return network.nodes.lightning.reduce((total, node) => {
+      const channels = nodes[node.name]?.channels || [];
+      return channels.reduce((sum, channel) => {
+        if (channel.pending || channel.status !== 'Open') return sum;
+        let balance = 0;
+        if (node.name === selectedName) {
+          balance = parseInt(channel.localBalance || '0', 10);
+        } else if (selfPubkey && channel.pubkey === selfPubkey) {
+          balance = parseInt(channel.remoteBalance || '0', 10);
+        }
+        return sum + (Number.isFinite(balance) ? balance : 0);
+      }, total);
+    }, 0);
+  }, [nodes, network.nodes.lightning, selectedName]);
+  const hasNoPaymentFunds =
+    assetId === 'sats' &&
+    !!selectedNode &&
+    !getAssetsAsync.loading &&
+    nodes[selectedName]?.channels !== undefined &&
+    outboundLiquidity <= 0;
 
   const payAsync = useAsyncCallback(async (values: FormValues) => {
     try {
@@ -104,6 +146,7 @@ const PayInvoiceModal: React.FC<Props> = ({ network }) => {
       okText={l('okBtn')}
       okButtonProps={{
         loading: payAsync.loading,
+        disabled: payAsync.loading || hasNoPaymentFunds,
       }}
       onOk={form.submit}
     >
@@ -119,6 +162,7 @@ const PayInvoiceModal: React.FC<Props> = ({ network }) => {
           onFinish={payAsync.execute}
           disabled={payAsync.loading}
         >
+          {hasNoPaymentFunds && <Alert type="warning" message={l('noFundsError')} />}
           <LightningNodeSelect
             network={network}
             name="node"
--- a/src/components/designer/lightning/actions/PayInvoiceModal.spec.tsx
+++ b/src/components/designer/lightning/actions/PayInvoiceModal.spec.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { act, fireEvent, waitFor } from '@testing-library/react';
 import { Status } from 'shared/types';
+
 import { LightningNodeChannelAsset } from 'lib/lightning/types';
 import { Network } from 'types';
 import { initChartFromNetwork } from 'utils/chart';
@@ -8,6 +9,7 @@ import { defaultRepoState } from 'utils/constants';
 import { createNetwork, mapToTapd } from 'utils/network';
 import {
   defaultStateChannel,
+  defaultStateInfo,
   getNetwork,
   lightningServiceMock,
   renderWithProviders,
@@ -23,6 +25,18 @@ describe('PayInvoiceModal', () => {
 
   beforeEach(() => {
     network = getNetwork(1, 'test network');
+    lightningServiceMock.getBalances.mockResolvedValue({
+      confirmed: '1000',
+      total: '1000',
+      unconfirmed: '0',
+    });
+    lightningServiceMock.getChannels.mockResolvedValue([
+      defaultStateChannel({
+        uniqueId: 'channel-1',
+        localBalance: '1000',
+        remoteBalance: '0',
+      }),
+    ]);
   });
 
   const renderComponent = async (nodeName = 'alice') => {
@@ -46,6 +60,9 @@ describe('PayInvoiceModal', () => {
     const cmp = <PayInvoiceModal network={network} />;
     const result = renderWithProviders(cmp, { initialState });
     unmount = result.unmount;
+    if (nodeName !== 'invalid') {
+      await result.findByLabelText('BOLT 11 Invoice');
+    }
     return result;
   };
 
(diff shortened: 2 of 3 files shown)


Abdulkbk:
Thanks for the pr @Leeyah-123. First pass, noticed the test coverage has dropped (you can verify locally `yarn ci`), specifically lines: 10, 17, 20 in `utils.ts`

image

Leeyah-123:
> Thanks for the pr @Leeyah-123. First pass, noticed the test coverage has dropped (you can verify locally `yarn ci`), specifically lines: 10, 17, 20 in `utils.ts`
> 
> image
Thanks for the feedback! Checking that out now.

src/lib/lightning/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { LightningNodeBalances, LightningNodeChannel } from './types';
+
+export const NO_LIGHTNING_PAYMENT_FUNDS_ERROR =
+  'Node has no funds available to pay invoices. Fund the node or open a channel with outbound liquidity and try again.';
```
Jem256: @Leeyah-123 Thanks for working on this! 
nit: The rest of the codebase uses `usePrefixedTranslation` for user-facing strings. Would it make sense for this message to go through the i18n system instead of being a hardcoded English string, for consistency?

Leeyah-123: Oh, I checked similar guard rails in the codebase, like in [src/store/models/lit.ts#L151-153](https://github.com/Leeyah-123/polar/blob/85d4eef5dd1646f0856e2f0dcb0fe19b29618e16/src/store/models/lit.ts#L151-L153), and saw that the errors are thrown in English, but it does make more sense to go through the i18n system instead. Working on this now, thank you!

src/store/models/lightning.ts
```diff
@@ -283,6 +288,15 @@ const lightningModel: LightningModel = {
       { injections, getStoreState, getStoreActions },
     ) => {
       const api = injections.lightningFactory.getService(node);
+      await actions.getWalletBalance(node);
+      const channels = await actions.getChannels(node);
+      const nodeState = getStoreState().lightning.nodes[node.name];
+      const confirmedBalance = getConfirmedLightningWalletBalance(
+        nodeState?.walletBalance,
+      );
+      if (confirmedBalance <= 0 && getOutboundLightningLiquidity(channels) <= 0) {
```
Jem256: The original NaN sats bug still reproduces on c-lightning nodes after this PR.
**Repro steps:**
1. Create a default network and start it.
2. Fund a CLN node (e.g. carol) with Actions → Deposit. Don't open any channels.
3. Create an invoice on any other node.
4. From Carol, Actions → Pay Invoice and paste the BOLT 11.
Click Pay Invoice.

The check is `confirmedBalance <= 0 && outboundLiquidity <= 0`. Since carol has on-chain funds from the deposit, `confirmedBalance > 0`, and the guard passes. 
Suggestion: for payments, only outbound-channel liquidity is usable. On-chain funds can't pay a Lightning invoice. Could the guard drop the confirmedBalance check entirely?

`if (getOutboundLightningLiquidity(channels) <= 0) {
  throw new Error(NO_LIGHTNING_PAYMENT_FUNDS_ERROR);
}`

src/components/designer/lightning/actions/PayInvoiceModal.tsx
```diff
@@ -48,7 +53,7 @@ const PayInvoiceModal: React.FC<Props> = ({ network }) => {
 
   const [form] = Form.useForm();
   const assetId = Form.useWatch<string>('assetId', form) || 'sats';
-  const selectedName = Form.useWatch<string>('node', form) || '';
+  const selectedName = Form.useWatch<string>('node', form) || nodeName || '';
 
   const getAssetsAsync = useAsync(async () => {
```
Jem256, line 60: Immediately after network creation, the proactive guard silently doesn't fire. There's a slight lag between initial render of the modal and when the warning is triggered. 
Would it work to also fetch balance and channels for the selected node here? refer [here](https://github.com/jamaljsr/polar/pull/1367#discussion_r3001776802) for details and suggested code

src/lib/lightning/utils.ts
Jem256: On second thought, I think your initial implementation of having it here in English may be a better approach. I had suggested the use of `usePrefixedTranslation` because you had a user-facing modal that would benefit from the translation. However, we now have two copies of the same string in different places, and having a single source of truth is a much better option in my opinion. Happy to defer to your judgment on this, though if you see it differently.

src/store/models/lightning.ts
```diff
@@ -283,6 +288,15 @@ const lightningModel: LightningModel = {
       { injections, getStoreState, getStoreActions },
     ) => {
       const api = injections.lightningFactory.getService(node);
+      await actions.getWalletBalance(node);
```
Jem256: `balanceChannels` (around L461 of this file) calls `payInvoice` in a `Promise.all` loop for every rebalancing payment. With the new guard, each of those calls now fires an extra `getWalletBalance` + `getChannels` per payment. For an N-channel rebalance, that's 2N extra concurrent API calls. 
Would it be worth splitting the funds check out so `balanceChannels` can bypass it? Either an opt-out param on `payInvoice`, or extracting an internal thunk that `balanceChannels` calls directly, whichever feels cleaner.

Jem256 (changes requested):
Thanks for the work on this, @Leeyah-123. The modal-level guard catches the original NaN case for LND, which is good. I tested the PR locally and found a few things worth addressing. 
The main one is that the original "Sent NaN sats" bug still reproduces on c-lightning nodes. The guard's logic treats on-chain balance as sufficient for invoice payments, but on-chain funds can't actually pay a Lightning invoice, so a CLN node that's been deposited to but has no channels still gets past the guard and produces the same broken success toast. Details inline.
Beyond that, just regurgitating the greptile-apps bot comments: the proactive UI warning doesn't activate for LND/CLN/Eclair on first open (the useAsync only fetches data for litd nodes), and `balanceChannels` now fires redundant balance/channels calls for every rebalancing payment.

Leeyah-123:
> Thanks for the work on this, @Leeyah-123. The modal-level guard catches the original NaN case for LND, which is good. I tested the PR locally and found a few things worth addressing. The main one is that the original "Sent NaN sats" bug still reproduces on c-lightning nodes. The guard's logic treats on-chain balance as sufficient for invoice payments, but on-chain funds can't actually pay a Lightning invoice, so a CLN node that's been deposited to but has no channels still gets past the guard and produces the same broken success toast. Details inline. Beyond that, just regurgitating the greptile-apps bot comments: the proactive UI warning doesn't activate for LND/CLN/Eclair on first open (the useAsync only fetches data for litd nodes), and `balanceChannels` now fires redundant balance/channels calls for every rebalancing payment.

Thanks a lot for the review @Jem256, working on fixing all you pointed out!

Leeyah-123:
> Thanks for the work on this, @Leeyah-123. The modal-level guard catches the original NaN case for LND, which is good. I tested the PR locally and found a few things worth addressing. The main one is that the original "Sent NaN sats" bug still reproduces on c-lightning nodes. The guard's logic treats on-chain balance as sufficient for invoice payments, but on-chain funds can't actually pay a Lightning invoice, so a CLN node that's been deposited to but has no channels still gets past the guard and produces the same broken success toast. Details inline. Beyond that, just regurgitating the greptile-apps bot comments: the proactive UI warning doesn't activate for LND/CLN/Eclair on first open (the useAsync only fetches data for litd nodes), and `balanceChannels` now fires redundant balance/channels calls for every rebalancing payment.


Hello @Jem256! Thanks a lot for the feedback, all reported issues are now fixed!

**Issue 1 – CLN `NaN sats` bug**
The root cause was the guard using `&&` with `confirmedBalance`, a node that had been deposited to but had no channels, would pass the check because `confirmedBalance > 0`, even though on-chain funds can't pay a Lightning invoice. Fixed by dropping the `confirmedBalance` check entirely from both the `payInvoice` action guard (`lightning.ts`) and the `hasNoPaymentFunds` flag in `PayInvoiceModal.tsx`. The guard now only looks at `getOutboundLightningLiquidity(channels) <= 0`.

**Issue 2 – Warning didn't fire on modal open for non-litd nodes**
The `getAssetsAsync` effect only looped over `litd` nodes, so when the modal opened against an LND/CLN/Eclair node, that node's `channels` were still `undefined` in the store. Because `hasNoPaymentFunds` gates on `channels !== undefined`, the warning was silently suppressed. Fixed by:
- Fetching `getWalletBalance` + `getChannels` for the initially selected node (regardless of implementation) at the start of `getAssetsAsync`, before the litd loop.
- Adding a secondary `useAsync` effect that re-runs balance + channel fetches whenever the selected node changes mid-modal, so switching nodes in the dropdown gives a warning state instantly.

**Issue 3 – Redundant API calls from `balanceChannels`**
Since `payInvoice` now fetches balance and channels as part of the funds guard, and `balanceChannels` calls `payInvoice` in a `Promise.all` loop, this would have produced 2N extra API calls per rebalance. Added `skipFundsCheck?: boolean` to `PayInvoicePayload` and wrapped the guard behind `if (!skipFundsCheck)`. `balanceChannels` passes `skipFundsCheck: true`, then it already has channel data loaded, and its payments always have sufficient liquidity by construction.

src/store/models/lightning.ts
```diff
@@ -279,10 +285,20 @@ const lightningModel: LightningModel = {
...
       const api = injections.lightningFactory.getService(node);
+
+      if (!skipFundsCheck) {
+        // Always refresh balance for UI reactivity.
+        await actions.getWalletBalance(node);
+        const channels = await actions.getChannels(node);
+        if (getOutboundLightningLiquidity(channels) <= 0) {
+          throw new Error(NO_LIGHTNING_PAYMENT_FUNDS_ERROR);
+        }
+      }
```
jamaljsr: I don't think it's necessary to also check here in the store when you already have the validation in the UI. I understand that the MCP tools work directly on the store but I think it's fine to make the LND call and return it's error directly.

src/lib/lightning/utils.ts
```diff
@@ -0,0 +1,15 @@
+import { LightningNodeChannel } from './types';
+
+export const NO_LIGHTNING_PAYMENT_FUNDS_ERROR =
+  'Node has no funds available to pay invoices. Fund the node or open a channel with outbound liquidity and try again.';
+
+export const getOutboundLightningLiquidity = (
```
jamaljsr: If the store validation is removed, I'd suggest moving this logic into the component instead of creating this one-off file. The const can be removed as well.

jamaljsr (commented):
@Leeyah-123 Thanks for this PR. I tested this and it works as advertised.

In general, I've been pretty relaxed on the UI validation because Polar is a developer tool, not really suited for non-technical folks. I believe it's generally better to display the errors directly from the nodes so they can use that message search for additional info if necessary. With that said, I'm ok with merging this PR since the scope is pretty small.

I have some minor suggestions on the code. Also, please rebase this and it should be good to go.

Leeyah-123:
> @Leeyah-123 Thanks for this PR. I tested this and it works as advertised.
> 
> In general, I've been pretty relaxed on the UI validation because Polar is a developer tool, not really suited for non-technical folks. I believe it's generally better to display the errors directly from the nodes so they can use that message search for additional info if necessary. With that said, I'm ok with merging this PR since the scope is pretty small.
> 
> I have some minor suggestions on the code. Also, please rebase this and it should be good to go.

Thanks a lot for the feedback @jamaljsr! I've implemented the suggestions and rebased this as well.

Abdulkbk:
@Leeyah-123 the pr includes 3 commits (from another PR) that are already merged in master. Could you fix that?.

Leeyah-123:
> @Leeyah-123 the pr includes 3 commits (from another PR) that are already merged in master. Could you fix that?.

@Abdulkbk Apologies for the oversight, it has been fixed!

Jem256 (commented):
LGTM
Tested on Linux and Windows.

cc. @Abdulkbk

src/store/models/lightning.ts
```diff
@@ -2,6 +2,7 @@ import { Action, action, Thunk, thunk, ThunkOn, thunkOn } from 'easy-peasy';
 import { throttle } from 'lodash';
```
Abdulkbk, line 2: Nothing in this file was changed in the end, only newlines were added. Can you revert it?

src/components/designer/lightning/actions/PayInvoiceModal.tsx
```diff
@@ -1,22 +1,23 @@
-import React, { useMemo } from 'react';
-import { useAsync, useAsyncCallback } from 'react-async-hook';
 import styled from '@emotion/styled';
```
Abdulkbk, line 1: I don't think this makes sense as a separate commit, the loader shouldn't have been removed in the first place. Maybe squash this commit into the previous one.

src/lib/lightning/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { LightningNodeBalances, LightningNodeChannel } from './types';
```
Abdulkbk: This file and `util.spec.ts` were later deleted in a new commit. I think we should edit the commits that introduced them and drop them there.

For example, this commit added `util.ts`. We should edit the commit and drop it here.

src/store/models/lightning.spec.ts
```diff
@@ -125,6 +126,35 @@ describe('Lightning Model', () => {
     expect(channels).toEqual([]);
   });
 
+  it('should pay invoices when the node has outbound liquidity', async () => {
```
Abdulkbk: Since the changes made to `lightning.ts` were later reversed, shouldn't we also reverse these tests?

src/lib/lightning/utils.spec.ts
```diff
@@ -0,0 +1,103 @@
+import { defaultStateChannel } from 'utils/tests';
```
Abdulkbk: Just drop this commit since the files was later deleted in a later commit.

Abdulkbk (changes requested):
Looking good.

I have a few pieces of feedback regarding the PR structure. Also rebase the branch.

Leeyah-123:
> Looking good.
> 
> I have a few pieces of feedback regarding the PR structure. Also rebase the branch.

@Abdulkbk On it!

Leeyah-123:
Thanks for the review, @Abdulkbk!
I've rebased onto the latest master and restructured the branch to address the feedback.
The history is now a single commit with all the necessary changes.

src/components/designer/lightning/actions/PayInvoiceModal.tsx
```diff
@@ -38,31 +39,69 @@ interface Props {
...
-  const selectedName = Form.useWatch<string>('node', form) || '';
+  const selectedName = Form.useWatch<string>('node', form) || nodeName || '';
 
   const getAssetsAsync = useAsync(async () => {
     if (!visible) return;
+
+    // Fetch balance + channels for the initially selected node
+    // so the proactive warning renders correctly on open.
+    if (selectedNode) {
+      await getWalletBalance(selectedNode);
```
Abdulkbk: does not seem like we're using wallet balance. If so we can remove the call here.

src/components/designer/lightning/actions/PayInvoiceModal.tsx
```diff
@@ -38,31 +39,69 @@ interface Props {
...
+      // avoid double-fetching channels for the already-selected litd node
+      if (node.name !== selectedNode?.name) await getChannels(node);
       await getAssetRoots(mapToTapd(node));
     }
   }, [network.nodes, visible]);
 
+  // Re-fetch balance + channels when the user switches the selected node mid-modal.
+  useAsync(async () => {
+    if (!visible || !selectedNode) return;
+    await getWalletBalance(selectedNode);
```
Abdulkbk: here as well, seems we use the channels purely to compute `outboundLiquidity`

src/components/designer/lightning/actions/PayInvoiceModal.tsx
```diff
@@ -38,31 +39,69 @@ interface Props {
 const PayInvoiceModal: React.FC<Props> = ({ network }) => {
   const { l } = usePrefixedTranslation('cmps.designer.lightning.actions.PayInvoiceModal');
   const { visible, nodeName } = useStoreState(s => s.modals.payInvoice);
+  const { nodes } = useStoreState(s => s.lightning);
   const { getAssetsInChannels } = useStoreState(s => s.lit);
   const { formatAssetAmount } = useStoreState(s => s.tap);
   const { hidePayInvoice } = useStoreActions(s => s.modals);
-  const { payInvoice, getChannels, getInfo } = useStoreActions(s => s.lightning);
+  const { payInvoice, getChannels, getInfo, getWalletBalance } = useStoreActions(
```
Abdulkbk: > here as well, seems we use the channels purely to compute outboundLiquidity


if the comments below are correct, we can safely remove `getWalletBalance` from here.

Abdulkbk (changes requested):
Thanks for making the changes @Leeyah-123. One last feedback to address and this is good to go.

Abdulkbk (approved):
LGTM

PR #1369, opened 2026-04-20
add CLN support on windows 

### LINKED ISSUE #1139: CLN on Windows
I see CLN is not available for windows as of yet. Is there a timeline for adding this support? Does the support have to do with anything with CLN, or is it just polar that doesn't support it on windows yet?

Jem256:
Closes #1139 

### Description

CLN crashes on Windows when its data directory is a Docker bind mount. The gossipd daemon fails to initialize gossip_store and exits with STATUS_FAIL_INTERNAL_ERROR, causing a crash loop. This is due to filesystem incompatibilities in Docker Desktop's bind mount layer on Windows (9p/virtiofs → NTFS).
This PR fixes the crash by using a named Docker volume for CLN's data directory on Windows instead of a host bind mount. Named volumes are stored on the Docker VM's internal ext4 filesystem, which CLN handles without issue. Linux and macOS are unaffected; they continue using bind mounts with full host visibility.

### Steps to Test

1. Run Polar on Windows
2. Create a network with a CLN node
3. Start the network and verify that the CLN node starts successfully without crashing and can perform all node operations.
4. Verify the CLN node can be renamed and start with its previous state intact, network deletion, and node removal work as intended.
5. Verify on macOS/Linux that CLN nodes still start and rename correctly (bind mount behaviour unchanged)

### Screenshots

image

image


src/lib/docker/nodeTemplates.ts
```diff
@@ -72,14 +72,18 @@ export const clightning = (
...
   command: trimInside(command),
   restart: 'always',
   volumes: [
-    `./volumes/${dockerConfigs['c-lightning'].volumeDirName}/${name}/${dockerConfigs['c-lightning'].dataDir}:/home/clightning/.lightning`,
+    // on windows, use a named volume so the CLN data stays on docker VM ext4 filesystem
+    namedVolumeName
+      ? `${namedVolumeName}:/home/clightning/.lightning`
+      : `./volumes/${dockerConfigs['c-lightning'].volumeDirName}/${name}/${dockerConfigs['c-lightning'].dataDir}:/home/clightning/.lightning`,
     `./volumes/${dockerConfigs['c-lightning'].volumeDirName}/${name}/${dockerConfigs['c-lightning'].apiDir}:/opt/c-lightning-rest/certs`,
   ],
```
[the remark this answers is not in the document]
Jem256: Polar talks to CLN via the built-in `CLNREST` plugin ( configured with `--clnrest-port=8080` in `constants.ts`, not the legacy c-lighting-rest service. The `/opt/c-lightning-rest/certs` bind mount is essentially unused and will not be affected by any potential low-level file operations that this PR is fixing. 

Worth noting: would it be ideal to raise a separate cleanup issue for this or retain it for backward compatibility with older CLN images that bundled c-lightning-rest? 

cc @Abdulkbk

src/lib/docker/dockerService.ts
```diff
@@ -552,6 +568,57 @@ class DockerService implements DockerLibrary {
...
+      if (!copySuccessful) {
+        try {
+          await docker.getVolume(newVolumeName).remove();
+        } catch (e) {
+          info(`Failed to clean up new volume ${newVolumeName}: ${e}`);
+        }
+      }
+    }
+
+    await docker.getVolume(newVolumeName).remove();
```
Abdulkbk: Renaming a CLN node doesn't seem to work because of this.

Abdulkbk (commented):
Nice work @Jem256.

I completed a first pass and have the following feedback:
1. I noticed that CLN refuses to start after a rename.
2. When you export a network and then import it, CLN doesn't start.

These are what I have for now, but I will have more once I look deeper. It also needs a rebase.

src/lib/docker/dockerService.ts
```diff
@@ -552,6 +568,56 @@ class DockerService implements DockerLibrary {
...
+   * @param network the network containing the node
+   * @param node the node to be renamed
+   * @param newName the new name for the node and directory
+   */
+  private async renameCLNVolume(network: Network, node: CLightningNode, newName: string) {
+    const docker = await getDocker();
+    const networkName = `polar-network-${network.id}`;
+    const oldContainerName = getContainerName(node);
+    const newContainerName = `polar-n${network.id}-${newName}`;
+    const oldVolumeName = `${networkName}_${oldContainerName}`;
```
Abdulkbk, line 664: Do we ever remove the oldVolume after successful copy?.

src/store/models/network.ts
```diff
@@ -13,6 +14,7 @@ import {
   TapdNode,
   TapNode,
 } from 'shared/types';
+import { clightningService } from 'lib/lightning/clightning';
```
Abdulkbk: Ideally, we should not import this directly as it bypasses the DI/injection pattern. We should rather go through injections/lightningFactory.

src/lib/lightning/clightning/clightningService.ts
```diff
@@ -333,6 +342,161 @@ export class CLightningService implements LightningService {
...
+    try {
+      helper = await docker.createContainer({
+        Image: image,
+        Entrypoint: ['/bin/sh', '-c'],
+        Cmd: [
+          'rm -f /dest/regtest/lightning-rpc && ' +
+            'cd /source && tar --exclude=lightning-rpc -cf - . | tar -xf - -C /dest',
+        ],
+        HostConfig: {
+          Binds: [`${volumeName}:/source:ro`, `${hostDataDir}:/dest`],
```
Abdulkbk: I think we'd need to normalize the path to `POSIX` format before passing it here.

Abdulkbk (changes requested):
Thanks for the changes. I believe we're getting there. I left some feedback after looking at the code. Additionally, the feedback I raised in my previous review on CLN not starting  when a network is imported (from a prev export) persists.

src/lib/docker/dockerService.ts
Jem256: Thanks. This has been handled.

Abdulkbk (approved):
LGTM 🚢

Needs a rebase

PR #1376, opened 2026-06-22
feat: detect LND wallet locked state

### LINKED ISSUE #470: Feature Request: Locking Nodes
The new lnd v0.13.0 version changed a lot in regards to unlocking nodes.
I would like to be able to test if unlocking in my application works by locking those nodes in Polar

Ideally if a node was selected in the graph, in the details panel there could be a button "lock/unlock". 
To avoid password hazzle a standard password that is always "polar" could be used, as this is solely for testing purposes anyway.

If this can already be acheived with Polar, then a short answer on how to do this would also be sufficient.

Jem256:
Closes #470

### Description

This PR implements wallet locked state detection for LND nodes. Currently, Polar starts all LND nodes with the `--noseedbackup` flag, which skips wallet initialization and makes the node immediately available. This PR lets users start LND nodes in a locked state by removing the `--noseedbackup` flag from the node's default command in  `advanced options` before the initial start.

### Steps to Test

1. Click "Create Network"
2. Create the network with LND nodes. 
3. Open the Advanced Options modal for a node, remove  `--noseedbackup` from the default command, and save.
4. Start the network.
5. Verify the LND node shows a "Locked" status badge instead of "Started" or "Error"
6. Open a terminal on a locked node and run `lncli state` to confirm it returns `NON_EXISTING.`

### Screenshots

image

### DIFF
--- a/electron/lnd/lndProxyServer.ts
+++ b/electron/lnd/lndProxyServer.ts
@@ -253,8 +253,84 @@ export const initLndProxy = (ipc: IpcMain) => {
 };
 
 /**
+ * Separate cache for unauthenticated State RPC clients (TLS only, no macaroon).
+ * Used to query wallet state before the wallet is unlocked/created.
+ */
+let noAuthRpcCache: {
+  [key: string]: LND.StateApi;
+} = {};
+
+const getNoAuthRpc = async (node: LndNode): Promise<LND.StateApi> => {
+  const { name, ports, paths, networkId } = node;
+  const id = `n${networkId}-${name}`;
+  if (!noAuthRpcCache[id]) {
+    const config: LND.LndClientOptions = {
+      socket: `127.0.0.1:${ports.grpc}`,
+      cert: (await readFile(paths.tlsCert)).toString('hex'),
+      macaroon: '',
+    };
+    noAuthRpcCache[id] = LND.StateApi.create(config);
+  }
+  return noAuthRpcCache[id];
+};
+
+/**
+ * Evict both caches for a node. Called when a WalletUnlocker RPC errors,
+ * since a stale cached connection is a likely cause.
+ */
+const evictRpcCache = (node: LndNode) => {
+  const { name, networkId } = node;
+  const id = `n${networkId}-${name}`;
+  delete noAuthRpcCache[id];
+  delete rpcCache[id];
+};
+
+const getState = async (args: { node: LndNode }): Promise<LND.GetStateResponse> => {
+  const rpc = await getNoAuthRpc(args.node);
+  return await rpc.getState();
+};
+
+const walletUnlockerListeners: {
+  [key: string]: (...args: any) => Promise<any>;
+} = {
+  [ipcChannels.getState]: getState,
+};
+
+/**
+ * Register IPC handlers for WalletUnlocker RPCs under the
+ * 'lnd-wallet-unlocker' prefix. Same pattern as initLndProxy
+ * but uses TLS-only gRPC clients (no macaroon).
+ */
+export const initLndWalletUnlockerProxy = (ipc: IpcMain) => {
+  debug('LndWalletUnlockerProxy: initialize');
+  Object.entries(walletUnlockerListeners).forEach(([channel, func]) => {
+    const requestChan = `lnd-${channel}-request`;
+
+    debug(`LndWalletUnlockerProxy: listening for ipc command "${channel}"`);
+    ipc.on(requestChan, async (event, ...args) => {
+      let uniqueChan = `lnd-${channel}-response`;
+      if (args && args[0] && args[0].replyTo) {
+        uniqueChan = args[0].replyTo;
+      }
+      try {
+        const result = await func(...args);
+        debug(`LndWalletUnlockerProxy: send response "${uniqueChan}"`, toJSON(result));
+        event.reply(uniqueChan, result);
+      } catch (err: any) {
+        debug(`LndWalletUnlockerProxy: send error "${uniqueChan}"`, err);
+        if (args && args[0] && args[0].node) {
+          evictRpcCache(args[0].node);
+        }
+        event.reply(uniqueChan, { err: err.message });
+      }
+    });
+  });
+};
+
+/**
  * Clears the cached rpc instances
  */
 export const clearLndProxyCache = () => {
   rpcCache = {};
+  noAuthRpcCache = {};
 };
--- a/src/lib/lightning/lnd/lndService.ts
+++ b/src/lib/lightning/lnd/lndService.ts
@@ -4,7 +4,7 @@ import { PendingChannel } from 'shared/lndDefaults';
 import { LightningNode, LndNode, OpenChannelOptions } from 'shared/types';
 import * as PLN from 'lib/lightning/types';
 import { LightningService } from 'types';
-import { waitFor } from 'utils/async';
+import { AbortWaitError, waitFor } from 'utils/async';
 import { lndProxyClient as proxy } from './';
 import { mapOpenChannel, mapPendingChannel } from './mappers';
 
@@ -218,6 +218,11 @@ class LndService implements LightningService {
     };
   }
 
+  async getWalletState(node: LightningNode): Promise<LND.WalletState> {
+    const res = await proxy.getState(this.cast(node));
+    return res.state;
+  }
+
   /**
    * Helper function to continually query the LND node until a successful
    * response is received or it times out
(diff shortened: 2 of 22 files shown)


Jem256:
cc: @Abdulkbk for review

Abdulkbk:
This seems like a feature likely to be used by a technical subset of users. If that is the case, why not just add the "Locked" status, allow opening the terminal in that state so the user can edit the prefilled default commands and remove `--noseedbackup`, and have LND start in the locked state? This way, LND would not deviate much from the other lightning implementations?. Just thinking out loud here...

Jem256:
> This seems like a feature likely to be used by a technical subset of users. If that is the case, why not just add the "Locked" status, allow opening the terminal in that state so the user can edit the prefilled default commands and remove `--noseedbackup`, and have LND start in the locked state? This way, LND would not deviate much from the other lightning implementations?. Just thinking out loud here...

For clarity, you mean do away with the checkbox entirely and let it be triggered later by removing the `--noseedbackup` flag from `advanced options` before the network starts, right?

Abdulkbk:
>For clarity, you mean do away with the checkbox entirely and let it be triggered later by removing the --noseedbackup flag from advanced options before the network starts, right?

Yes, then ensure we handle the "Locked" state and do not mark the LND node as failed, since the current arch requires the node RPC to be active within 2 mins or so before polar marks it as failed. Additionally, the user should be able to access the terminal if the node is in the "Locked" state. But I might be missing something...

Jem256:
> Yes, then ensure we handle the "Locked" state and do not mark the LND node as failed, since the current arch requires the node RPC to be active within 2 mins or so before polar marks it as failed. Additionally, the user should be able to access the terminal if the node is in the "Locked" state. But I might be missing something...

The terminal is currently accessible for Locked nodes. Let me work on the suggestions.

src/shared/types.ts
```diff
@@ -4,6 +4,7 @@ export enum Status {
   Stopping,
   Stopped,
   Error,
+  Locked,
```
Abdulkbk, line 7: Adding `Locked` means a node can now be *running* without being `Status.Started`, and I think `getOpenPorts` in `src/utils/network.ts` needs to know about that.

All six filters there are `n.status !== Status.Started` (lines 793, 840, 873, 903, 925, 966), with a comment above them explaining why: started nodes are holding their own ports. A locked node is running with published ports but doesn't match that check, so `detectPort` sees its own port as taken, returns a different one, and we write the new ports into `network.json` and the compose file while the container keeps listening on the old ones.

I think you can hit this by starting any other stopped node while a locked node is up — have you tried that flow? Would a small `isNodeRunning(status)` helper covering `Started | Locked` work here, so the next status we add doesn't have to remember all six spots?

src/lib/lightning/lnd/lndService.ts
```diff
@@ -229,6 +257,16 @@ class LndService implements LightningService {
   ): Promise<void> {
     return waitFor(
       async () => {
+        const state = await this.getWalletState(node);
+        if (state === 'LOCKED' || state === 'NON_EXISTING') {
```
Abdulkbk: `LOCKED` is unambiguous, but I'm less sure about `NON_EXISTING`. LND registers the State service before creating the wallet, so a node running with `--noseedbackup` also briefly reports `NON_EXISTING` at startup until the wallet is auto-created.

`waitFor` polls at t=0 and then every 3s, so that window may well fall between polls most of the time. But if a poll does land in it, we abort and mark the healthy default node `Locked`, with no recovery (see my comment on `monitorStartup`). Did you see this at all while testing on a larger network, or on a slower machine?

Since the node's Docker command drives this feature, could we make it deterministic, treat `NON_EXISTING` as terminal only when the command explicitly lacks `--noseedbackup`? Requiring two consecutive `NON_EXISTING` reads would work too.

electron/lnd/lndProxyServer.ts
```diff
@@ -252,9 +252,113 @@ export const initLndProxy = (ipc: IpcMain) => {
...
 
+/**
+ * Separate cache for unauthenticated gRPC clients (TLS only, no macaroon).
+ * Used for State and WalletUnlocker RPCs that are available before wallet init.
+ */
+let noAuthRpcCache: {
+  [key: string]: LND.LndRpcApis;
+} = {};
+
+const getNoAuthRpc = async (node: LndNode): Promise<LND.LndRpcApis> => {
```
Abdulkbk: `LndClient.create` builds all 15 sub-server APIs — `lightning`, `router`, `signer`, `walletKit`, and the rest — each loading its own proto and opening its own gRPC channel. Here we only need `state` and `walletUnlocker`, and the result is cached for the process lifetime, so every LND node ends up holding a second full set of connections created with an empty macaroon.

Any reason not to use `LND.StateApi.create(config)` instead (plus `WalletUnlockerApi` if the wallet RPCs stay)?

electron/lnd/lndProxyServer.ts
```diff
@@ -252,9 +252,113 @@ export const initLndProxy = (ipc: IpcMain) => {
...
+  delete noAuthRpcCache[id];
+  delete rpcCache[id];
+};
+
+const getState = async (args: { node: LndNode }): Promise<LND.GetStateResponse> => {
+  const rpc = await getNoAuthRpc(args.node);
+  return await rpc.state.getState();
+};
+
+const genSeed = async (args: { node: LndNode }): Promise<LND.GenSeedResponse> => {
```
Abdulkbk: `genSeed`, `initWallet`, and `unlockWallet` are wired through the channel list, proxy server, proxy client, and service, but I couldn't find anything that calls themgrep only turns up the definitions. Intentional groundwork for the unlock UI?

If that UI is a separate PR, would you mind moving these three there? It'd keep this one scoped to the detection the title describes, and it avoids shipping an `initWallet` handler that writes wallet state with nothing on the other end.

electron/lnd/lndProxyServer.ts
```diff
@@ -252,9 +252,113 @@ export const initLndProxy = (ipc: IpcMain) => {
...
+
+    debug(`LndWalletUnlockerProxy: listening for ipc command "${channel}"`);
+    ipc.on(requestChan, async (event, ...args) => {
+      let uniqueChan = `lnd-${channel}-response`;
+      if (args && args[0] && args[0].replyTo) {
+        uniqueChan = args[0].replyTo;
+      }
+      try {
+        const result = await func(...args);
+        debug(`LndWalletUnlockerProxy: send response "${uniqueChan}"`, toJSON(result));
```
Abdulkbk, line 317: This `debug` runs for every handler in the map, so `genSeed` would log the full cipher seed mnemonic and `initWallet` the admin macaroon into `main.log` the file people tend to paste into bug reports.

Nothing calls those handlers today, so it isn't live, but it's easy to miss later. Could we skip or redact the payload for those two channels and keep the full log just for `getState`?

Abdulkbk (changes requested):
Nice work, left some few comments

Abdulkbk (changes requested):
Thanks for working on this. I have 2 comments, though I don't have the solutions for them:

1. When you started LND in locked state, and then use the terminal to unlock the node, it does not reflect on the UI. Although you can use the terminal to do all sort of things, the UI remains in locked state, preventing you from doing all kinds of operations.
2. When you start in normal mode, then stop the network, and restart in locked mode, although you can aqccess the terminal, but you won't be able to unlock/create a new wallet.

Kindly investigate these and let me know if you have an explanation/fix.

Jem256:
> Thanks for working on this. I have 2 comments, though I don't have the solutions for them:
> 
> 1. When you started LND in locked state, and then use the terminal to unlock the node, it does not reflect on the UI. Although you can use the terminal to do all sort of things, the UI remains in locked state, preventing you from doing all kinds of operations.
> 2. When you start in normal mode, then stop the network, and restart in locked mode, although you can aqccess the terminal, but you won't be able to unlock/create a new wallet.
> 
> Kindly investigate these and let me know if you have an explanation/fix.

@Abdulkbk thank you for the comments. 

1. I scoped this PR to only cater for detecting the locked state. The follow-up [PR](https://github.com/jamaljsr/polar/pull/1384) that has wallet initialisation and unlock enables the user to create a wallet and unlock the node using both the UI and terminal. 
2. When a node starts with the `--noseedbackup` flag, LND creates the wallet itself and encrypts it with two different default passphrases: `public` for the master public key and `hello` for the private one. Both the passphrases don't work with the `unlock` RPC since it only takes one password. The quick way around this default behavior is to call `changepassword`, leave the current password field empty (LND assumes this action as coming from a `--noseedbackup` state and re-encrypts with both defaults for you and then you can specify your own password to use for unlocking that node. However, I did not consider this flow in my wallet init & unlock PR. If it is something we should add on the UI side, please let me know

Abdulkbk (approved):
lgtm

PR #1380, opened 2026-07-16
test: cover error paths in the pay invoice modal

Abdulkbk:
Follow-up to #1367

### Description

`PayInvoiceModal.tsx` was at 96.51% statements / 89.28% functions. The gaps were all error paths introduced by #1367 that nothing exercised: 

- The three `.catch(() => undefined)` handlers around `getInfo`/`getChannels`. These exist to prevent an unreachable node from stopping the modal from opening — worth noting, since the modal fetches from every node in the network. One test covers the plain case; a second covers the litd loop's separate catch, which runs only for a litd node that isn't the selected one. 
- The `remoteBalance || '0'` fallback for a peer-opened channel where the selected node holds nothing. Now asserted to read as zero liquidity.

### Steps to Test
```bash
yarn ci
```

### DIFF
--- a/src/components/designer/lightning/actions/PayInvoiceModal.spec.tsx
+++ b/src/components/designer/lightning/actions/PayInvoiceModal.spec.tsx
@@ -200,6 +200,46 @@ describe('PayInvoiceModal', () => {
       ).not.toBeInTheDocument();
     });
 
+    it('should treat a peer-opened channel with no remote balance as no liquidity', async () => {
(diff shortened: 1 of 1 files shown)


Abdulkbk:
cc: @Jem256

Jem256 (commented):
Thank you for adding the coverage @Abdulkbk It does exactly what it was intended to do. This LGTM

Jem256 (commented):
Thank you for adding the coverage @Abdulkbk It does exactly what it was intended to do. This LGTM

PR #1382, opened 2026-07-23
fix(terminal): terminal fails to open with custom node

### LINKED ISSUE #1359: Bug: terminal fails to open with LND custom node
**Describe the bug**
When using a custom-built LND image in a Polar network, the terminal window (launched via the "Terminal" button in the UI) opens for a fraction of a second and immediately closes. This prevents any CLI interaction with the custom node.

**To Reproduce**
1. Build a custom LND Docker image (e.g., using a custom Dockerfile with specific user/group configurations).
2. Add the custom image to Polar via Settings > Docker Images.
3. Create a new network and add the custom LND node.
4. Start the network.
5. Click on the node and click the Terminal tab or button.
The terminal window pops up and crashes/exits instantly.

**Expected behavior**
The terminal should remain open and provide a functional shell session inside the running container.

- OS: macOs 15.2
- Polar Version: 4.0.0
- Docker Version: 28.0.4
- Docker Compose Version: 2.34.0

**Additional context**
I haven't tried this with other node implementations aside LND.

The Dockerfile used can be found [here](https://github.com/lightningnetwork/lnd/blob/master/Dockerfile).

Jem256:
Closes #1359

### Description

When using a custom-built LND image in a Polar network, the terminal window opens for a fraction of a second and immediately closes. The cause is two hardcoded assumptions in `DockerTerminal.tsx`: the exec requires `/bin/bash` and a user named `lnd` to exist in the container. Polar's managed images satisfy both via their entrypoint scripts, but custom images don't. So the exec fails silently, and the window closes before anyone can see why.

This PR tries the expected user and shell first, then falls back to the image's default user and to `/bin/sh` if needed, printing a note in the terminal whenever it falls back. If every attempt fails, the window now stays open and shows the error instead of closing.

### Steps to Test

1. Build a custom LND Docker image 
2. Add the custom image to Polar via Managed Images > Add a Custom Node.
3. Create a new network and add the custom LND node.
4. Start the network.
5. Click on the node and click the Terminal tab.

The terminal window should stay open, show a note that it's connected as the image's default user, and give you a working shell.

### Screenshots

image


src/components/terminal/DockerTerminal.tsx
```diff
@@ -57,19 +57,23 @@ U |  _"\\ u  \\/"_ \\/  |"|   U  /"\\  uU |  _"\\ u
 const nodeConfig: Record<string, { user: string; commands: string[] }> = {
   LND: {
     user: 'lnd',
-    commands: ['alias lncli="lncli --network regtest"'],
+    commands: ['alias lncli="lncli --network regtest --lnddir=/home/lnd/.lnd"'],
```
Abdulkbk, line 62: This makes sense, but it's only applied to 3 of 6 node types. `tapd` (`:80`) still uses `~/.tapd` and `litd` (`:86-87`) still uses `~/.lit`, any reason why?

src/components/terminal/DockerTerminal.tsx
```diff
@@ -105,11 +109,66 @@ const connectStreams = async (term: Terminal, name: string, type: string, l: any
...
+  // Custom images may not contain the user or the shell that Polar's managed
+  // images provide. Attempt a series of fallbacks instead.
+  const attempts = [
+    { user: config.user, shell: '/bin/bash' },
+    { user: config.user, shell: '/bin/sh' },
+    { user: undefined, shell: '/bin/bash' },
+    { user: undefined, shell: '/bin/sh' },
+  ];
+
+  let exec: any;
```
Abdulkbk: `exec` was `Docker.Exec` and `stream` a `Duplex` before this change; now `exec.resize`, `stream.write`, `s.destroy()` and `execState.ExitCode` are all unchecked. `@typescript-eslint/no-explicit-any` is `0` in `.eslintrc`, so CI won't flag it.

`let exec: Dockerode.Exec | undefined` / `let stream: Duplex | undefined` narrows fine with the existing break-then-null-check, and `Dockerode` is already imported in renderer-reachable code (`dockerService.ts:6`).

src/components/terminal/DockerTerminal.tsx
```diff
@@ -105,11 +109,66 @@ const connectStreams = async (term: Terminal, name: string, type: string, l: any
...
+        ...execCommand,
+        Cmd: [attempt.shell],
+        User: attempt.user,
+      });
+      const s = await e.start(execOptions);
+      // a missing shell binary may not throw on start with a TTY. give the
+      // shell a moment to fail fast, then confirm it is still running
+      await delay(100);
+      const execState = await e.inspect();
+      if (!execState.Running && execState.ExitCode !== 0) {
```
Abdulkbk: `ExecInspectInfo.ExitCode` is typed `number | null`. A shell that started and cleanly exited within the probe window passes this check, and we then `resize()` and write aliases into a dead exec. `if (!execState.Running)` is the condition you want — keep `ExitCode` for the message text.

src/components/terminal/DockerTerminal.tsx
```diff
@@ -105,11 +109,66 @@ const connectStreams = async (term: Terminal, name: string, type: string, l: any
...
+        s.destroy();
```
Abdulkbk: `s.destroy()` only runs on this branch. If `inspect()` on the previous line rejects, control jumps to `catch` with the hijacked socket still open. Narrow case — `start()` must have upgraded successfully first — but easy to cover by destroying in the `catch` as well.

src/components/terminal/DockerTerminal.tsx
```diff
@@ -105,11 +109,66 @@ const connectStreams = async (term: Terminal, name: string, type: string, l: any
...
+
+  if (!exec || !stream || !connected) {
+    // write the errors to the terminal so the user can read them
+    term.writeln(red(l('execErr')));
+    errors.forEach(e => term.writeln(red(`  - ${e}`)));
+    throw new Error(l('connectErr'));
+  }
+
+  // inform the user when a fallback was used
+  if (connected.user !== config.user) {
```
Abdulkbk, line 166: These assert a cause that wasn't actually verified — the attempt could have failed on a permission or daemon error rather than a missing user/shell. Something like "could not start a shell as '{{user}}' — connected as the image's default user instead" is more honest, and the specific reasons are already in `errors[]` if you want to print them alongside.

Abdulkbk (commented):
Reviewed and left a few feedback

PR #1384, opened 2026-07-31
feat: LND wallet initialization and unlock

### LINKED ISSUE #470: Feature Request: Locking Nodes

Jem256:
Closes #470

### Description

This PR implements wallet initialization and unlocking for locked LND nodes. It is a follow-up PR to #1376 and lets users initialize a wallet by setting a password; a seed phrase is generated for them before starting the node. Initialized nodes can also be unlocked through the UI. 

### Steps to Test

1. Create a network with an LND node, open Advanced Options, and remove the --noseedbackup startup flag so it boots locked.
2. Start the network and confirm the node shows the Locked badge/tag instead of an error.
3. Right-click the node (or open its details panel) and click Unlock.
4. If no wallet exists, confirm the initialize flow: enter a password (min 8 chars), submit, verify a seed phrase is shown and copyable.
5. Close the modal and confirm the node transitions to Started.
6. Restart the same node and confirm the unlock flow (password only, no seed) also transitions it to Started.
7. While locked, open the Connect tab and confirm it shows cert/host/port info with an explanatory alert.

### Screenshots

image

image

image


Abdulkbk (approved):

Abdulkbk (changes requested):
rebase and fix conflicts

src/components/common/UnlockNodeModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+      }
+      cancelText={l('cancelBtn')}
+      okText={l(notInitialized ? 'initOkBtn' : 'okBtn')}
+      okButtonProps={{
+        loading: submitAsync.loading,
+        disabled: walletStateAsync.loading || !!walletStateError,
+      }}
+      onOk={form.submit}
+    >
+      {walletStateAsync.loading ? (
```
Abdulkbk: The `loading`/`error` branches sit above the `mnemonic` branch, and `walletStateAsync` re-runs whenever `node`'s identity changes which `monitorStartup` guarantees right before `setMnemonic`. If that re-query is slow the seed is a spinner, and if it rejects the seed is gone for good. Should the `mnemonic` branch come first, and should the query short-circuit once we have a seed (`if (!visible || !node || mnemonic) return undefined`)?

src/components/common/UnlockNodeModal.tsx
```diff
@@ -0,0 +1,170 @@
...
+        const words = await initNode({ node, password });
+        setMnemonic(words);
+        notify({ message: l('initSuccess', { name: node.name }) });
+      } else {
+        await unlockNode({ node, password });
+        hideUnlockNode();
+        notify({ message: l('success', { name: node.name }) });
+      }
+    } catch (error: any) {
+      notify({ message: l(notInitialized ? 'initError' : 'error'), error });
```
Abdulkbk: `notInitialized` is re-read here after the `await`, but the state query can resolve in the meantime should we capture it at submit time so an init failure can't report "Unable to unlock the wallet"?

src/store/models/network.ts
```diff
@@ -881,6 +896,28 @@ const networkModel: NetworkModel = {
       const network = getStoreState().network.networks.find(n => n.id === id);
       if (!network) throw new Error(l('networkByIdErr', { networkId: id }));
 
+      // once a node is found to be Locked, keep polling its wallet state in the
+      // background so the UI notices if it gets unlocked outside of Polar
+      const pollForExternalUnlock = (ln: LndNode) => {
```
Abdulkbk, line 903: Every `monitorStartup` that ends in `Locked` starts another interval, and each one only clears when it observes the status change why not key the timers in a module-level map and clear the existing one first, so a second poller can't stack on a node that stayed locked?

src/store/models/network.ts
```diff
@@ -1186,6 +1224,16 @@ const networkModel: NetworkModel = {
       }
     },
   ),
+  initNode: thunk(async (actions, { node, password }, { injections }) => {
```
Abdulkbk, line 1234: Any reason the payload is `LightningNode` with three `as LndNode` casts rather than just typing it `LndNode`? The button is rendered off a bare `status === Locked` check, so nothing structural keeps a CLN node out.

src/components/designer/lightning/ConnectTab.tsx
```diff
@@ -83,27 +98,43 @@ const ConnectTab: React.FC<Props> = ({ node }) => {
...
+  const { getWalletState } = useStoreActions(s => s.lightning);
   const nodeState = useStoreState(s => s.lightning.nodes[node.name]);
   const pubkey = nodeState && nodeState.info ? nodeState.info.pubkey : '';
   const p2pLnUrlInternal = nodeState && nodeState.info ? nodeState.info.rpcUrl : '';
 
+  // a Locked LND node could be waiting to be unlocked (macaroons on disk from a
+  // previous init) or waiting to be initialized (no macaroons yet)
+  // Status.Locked alone doesn't tell us which, so query the wallet state directly
+  const isLockedLnd = node.status === Status.Locked && node.implementation === 'LND';
+  const walletStateAsync = useAsync(async (): Promise<LND.WalletState | undefined> => {
```
Abdulkbk, line 110: The callback closes over `node` but the deps are `[node.name, isLockedLnd]` should `node` (or `node.ports.grpc`) be in there too, so a port reassignment doesn't leave us querying with a stale node?

Abdulkbk (commented):
Read through the wallet init/unlock flow, the shape of it looks right (proxy pattern, injected `lndService`, error paths covered in tests). A few questions inline, mostly around the seed display and the unlock poller.

src/components/designer/NodeContextMenu.tsx
```diff
@@ -96,6 +97,11 @@ const NodeContextMenu: React.FC<Props> = ({ node: { id }, children }) => {
       [Status.Started, Status.Locked].includes(node.status),
     ),
     isStarted ? [{ type: 'divider' }] : [],
+    addItemIf(
```
Abdulkbk, line 100: The divider just above is gated on `isStarted`, so a locked node gets "Unlock" with no separator should that be `isStarted || node.status === Status.Locked`, matching the Terminal and Stop items?

electron/lnd/lndProxyServer.ts
```diff
@@ -314,7 +374,11 @@ export const initLndWalletUnlockerProxy = (ipc: IpcMain) => {
       }
       try {
         const result = await func(...args);
-        debug(`LndWalletUnlockerProxy: send response "${uniqueChan}"`, toJSON(result));
+        if (REDACTED_RESPONSE_CHANNELS.has(channel)) {
```
Abdulkbk: `createIpcSender` already logs the full request and response payload one call earlier, and `stripNode` only strips `node` so the mnemonic and admin macaroon reach the log anyway. Should the same channel set be redacted in `createIpcSender` (both directions), or should this be dropped since the comment promises something it doesn't deliver?

Abdulkbk (changes requested):

src/store/models/network.ts
```diff
@@ -1186,6 +1231,16 @@ const networkModel: NetworkModel = {
...
+    const mnemonic = await injections.lndService.genSeed(node);
+    await injections.lndService.initWallet(node, password, mnemonic);
+    await actions.monitorStartup([node]);
+    return mnemonic;
+  }),
+  unlockNode: thunk(async (actions, { node, password }, { injections }) => {
```
Abdulkbk, line 1260: One observation, otherwise, good to go.

While the unlock is in flight the node is still `Locked`, so the poller keeps ticking. If its 3s tick lands before this `monitorStartup` sets `Started`, it sees `RPC_ACTIVE` and fires a second `monitorStartup`. Both then run the tail at line 1012, so `addListeners(network)` runs twice and `subscribeChannelEvents` opens a second gRPC stream per node (the client overwrites `this.listeners[channel]`, so the first one can never be unsubscribed).

Since `unlockNode`/`initNode` own the resume themselves, should they clear the node's entry in `unlockPollers` before calling `monitorStartup`,  leaving the poller for genuinely external unlocks only?

Abdulkbk (approved):
LGTM

PR #1386, opened 2026-08-12
chore(deps): add Dependabot config for npm and GitHub Actions

HACKWITHNESBITT:
## Summary 

Add Dependabot configuration to automatically open dependency update PRs weekly for npm and GitHub Actions.

## Why 

Keeps dependencies current and reduces bit-rot.
Safe: Dependabot only opens PRs for maintainers to review; no code or runtime changes are applied automatically.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(deps)"
+    # Do NOT enable automerge here to keep zero-risk: maintainers review updates before merging.
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(deps)"


Jem256:
Hey @HACKWITHNESBITT, is this meant to replace Renovate or run alongside it? The repo has a [`.renovaterc`](https://github.com/jamaljsr/polar/blob/master/.renovaterc) at the root that already covers npm; it groups non-major updates and auto-merges them via branch automerge, with majors gated behind dependency dashboard approval. renovate[bot] has open PRs from that config (e.g. #1352, #1350). It covers GitHub Actions too, via the default manager set.

If both run, wouldn't we end up with two bots opening PRs for the same upgrades and competing `yarn.lock` rewrites? Or is there something specific Renovate is missing here?

HACKWITHNESBITT:
Thanks for the catch @Jem256,  good question.

I didn't intend to conflict with the existing Renovate setup. I opened this to add Dependabot checks, but I see now that Renovate is already handling npm (and Actions). I can take one of these approaches  please let me know which you prefer:

1) Keep Renovate as-is and limit this PR so Dependabot only manages GitHub Actions (no npm), to avoid duplicate PRs and competing lockfile updates.
2) Replace Renovate with Dependabot for npm + Actions (I'd remove/disable .renovaterc and confirm merge strategy for majors vs auto-merge).
3) Keep both but scope them so they don't overlap — e.g., Dependabot only for Actions and Renovate for npm, or point Dependabot to additional subdirectories if you want it to cover the nested example project.

If you prefer option (1), I’ll update this PR to only enable GitHub Actions updates. If you prefer (2) or (3), tell me which and I’ll update the config accordingly.


Why this matters

1. Running Renovate and Dependabot over the same ecosystems can produce duplicate PRs and competing lockfile changes that break CI and make review harder. Picking a single manager per ecosystem (or scoping them so they don't overlap) avoids churn.
2. The Greptile note about the nested npm project is important: the current Dependabot entry scans only /, so the example's package.json at examples/typescript/lnd is excluded unless you add another npm entry.

PR #1393, opened 2026-09-05
Restore an LND node from seedphrase 

### LINKED ISSUE #635: Seed phrases and channels backup
I like Polar and the testnet. Lately I decided to play around with backup/restore of nodes but have not access to seed phrase / channels backups?

Would be nice if the seed phrase could be displayed somewhere; as well as the option to start a new bitcoin node using an existing seed phrase.
Similarly for the Lightning Network: "backup channels" button + "restore channels" option.

Many people don't test their recovery strategy, would be nice to get familiarised with this the easy way using Polar 💁‍♂️

Jem256:
Closes #635 

### Description

This PR implements restoring a wallet from an existing seed phrase for uninitialized LND nodes. It is a follow-up PR to #1384 and lets users bring back a wallet they already have by pasting its 24-word seed, optionally attaching a channel backup file at the same time; both are sent as sibling fields on a single InitWallet request. Restores set a recovery window so LND scans for the seed's on-chain funds, and once a backup is attached, the peers force-close existing channels. 

Note: seeds carrying an aezeed passphrase are not supported in this PR.

### Steps to Test

1. Create a network, edit Alice's advanced options to remove --noseedbackup, and start.
2. Open Alice → Unlock → create wallet. Copy the seed; it is only shown once.
3. Deposit funds to Alice, open a channel to Bob, mine to confirm.
4. Copy the `channel.backup` file out of Alice's volume path.
5. Stop Alice. Delete volumes/lnd/alice/data/
6. Start Alice. She should come back `Locked` with wallet state `NON_EXISTING`.
7. Unlock → Restore → paste the seed, attach the backup, submit.
8. Confirm Alice's on-chain balance reappears and the channel shows as Force Closing rather than open.
9. Mine a few more blocks and confirm Alice's channel balance lands back in her wallet.
10. Repeat the restore on a second node without attaching a backup file, and confirm only on-chain balance is restored with no channel activity.

### Screenshots

image


src/lib/lightning/lnd/lndService.ts
```diff
@@ -276,8 +295,9 @@ class LndService implements LightningService {
           throw new AbortWaitError('wallet-not-initialized');
         }
         nonExistingCount = 0;
-        if (state !== 'RPC_ACTIVE' && state !== 'SERVER_ACTIVE') {
-          throw new Error(`waiting for RPC_ACTIVE, current state: ${state}`);
+        // RPC_ACTIVE precedes the p2p server starting, so wait for SERVER_ACTIVE
+        if (state !== 'SERVER_ACTIVE') {
```
Abdulkbk: This tightens the online check from `RPC_ACTIVE || SERVER_ACTIVE` to `SERVER_ACTIVE` only, which applies to **every** LND node in every network rather than just restored ones — was the wider scope intended, or is it needed specifically so peers are connectable for the SCB restore?

What I'm wondering about: `SERVER_ACTIVE` waits on chain sync and the p2p server, so a node that reaches `RPC_ACTIVE` and then stalls — an unsynced backend, or the `recoveryWindow: 2500` rescan this PR introduces on a longer chain — now burns the full 120s timeout. `monitorStartup` turns that into `Status.Error`, and unlike the `Locked` path there's no recovery poller, so the node stays in Error until the user restarts it. Previously it would have been marked `Started`.

If it is intended, could we get a line in the PR description flagging it? It's a 4-line hunk with a wide blast radius and it's easy to miss in a feature PR. And would it be worth a longer `waitUntilOnline` timeout for the restore path specifically, given the rescan is a known added delay?

src/store/models/network.ts
```diff
@@ -1268,6 +1285,72 @@ const networkModel: NetworkModel = {
...
+
+        const network = getStoreState().network.networkById(node.networkId);
+        const backend = network.nodes.bitcoin.find(n => n.name === node.backendName);
+        if (!backend || backend.status !== Status.Started) return;
+
+        // peers force-close after data-loss-protection, so confirm in the background
+        const confirmForceCloses = async () => {
+          const btcApi = injections.bitcoinFactory.getService(backend);
+          // closing txids are unknown until confirmed, so watch the funding outputs
+          const points = await injections.lndService.getRecoveredChannelPoints(node);
```
Abdulkbk: Is `waitingCloseChannels` guaranteed to be populated by the time `monitorStartup` resolves? My understanding is that the SCB is processed during server startup, so it's *usually* there by `SERVER_ACTIVE` — but if this lands early and returns `[]`, the `if (!points.length) return` below short-circuits the whole confirm-and-mine step, and the user has already been shown "Its channels are being force-closed. Mine more blocks to see the funds return." The channels then sit pending until they mine manually, with nothing in the log explaining why.

Would it be worth retrying the lookup for a few seconds so an early empty response doesn't silently skip the step?

```ts
const points = await waitFor(
  async () => {
    const cps = await injections.lndService.getRecoveredChannelPoints(node);
    if (!cps.length) throw new Error('waiting for recovered channels');
    return cps;
  },
  500,
  5 * 1000,
).catch(() => [] as string[]);
if (!points.length) {
  info(`No recovered channels found for '${node.name}' despite a channel backup`);
  return;
}
```

Abdulkbk (commented):
Nice work, left a questions

src/store/models/network.ts
```diff
@@ -1268,6 +1285,72 @@ const networkModel: NetworkModel = {
...
+              1000,
+              FORCE_CLOSE_WAIT_TIMEOUT,
+            );
+          } catch {
+            info(`Not all channels were force-closed for '${node.name}'`);
+          }
+          // mine even if some peers never closed, so the ones that did confirm
+          if ((await countSpent()) > 0) {
+            await getStoreActions().bitcoin.mine({ blocks: 1, node: backend });
+          }
```
Abdulkbk: Two small things here, both because `confirmForceCloses` is detached and can run for ~30s after the thunk resolves:

Should the backend's status be re-checked at this point? The `backend.status !== Status.Started` guard above runs before this task spawns, and `backend` is a by-value snapshot, so if the user stops the network or the backend right after the restore — plausible, since the modal has already closed — this fires `mine` at a stopped container. The outer `.catch` contains it, so it's log noise rather than breakage, but it looks cheap to avoid:

```suggestion
          // mine even if some peers never closed, so the ones that did confirm
          const latest = getStoreState().network.networkById(node.networkId);
          const live = latest.nodes.bitcoin.find(n => n.name === backend.name);
          if (live?.status === Status.Started && (await countSpent()) > 0) {
            await getStoreActions().bitcoin.mine({ blocks: 1, node: backend });
          }
```

Also — on the success path, `waitFor` has just established that every outpoint is spent, and then `countSpent()` re-issues a `gettxout` per channel to recompute the same number. Since `waitFor` resolves with the condition's return value, could the count be threaded through instead, leaving only the timeout branch to re-query?

src/components/common/RestoreWalletForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+  `,
+  Alert: styled(Alert)`
+    margin-bottom: 16px;
+  `,
+};
+
+const SEED_WORD_COUNT = 24;
+
+export const splitSeedWords = (phrase: string): string[] =>
+  phrase.trim().split(/\s+/).filter(Boolean);
```
Abdulkbk: Should this normalize the pasted phrase a bit more? Two cases I think users will hit:

The seed grid in `UnlockNodeModal` renders each word as `{i + 1}. {word}`. "Copy All" uses `mnemonic.join(' ')` and round-trips cleanly, but a user who selects and copies the grid itself gets `1. abandon 2. ability ...` — 48 tokens — and hits "must contain exactly 24 words" with no hint that the numbering is the problem. And a phrase that picked up capitalization somewhere passes the 24-word check, then fails at the RPC with an aezeed error that's much harder to act on.

```suggestion
export const splitSeedWords = (phrase: string): string[] =>
  phrase
    .toLowerCase()
    .replace(/\b\d+[.)]\s*/g, '')
    .trim()
    .split(/\s+/)
    .filter(Boolean);
```

The existing whitespace test would still pass; worth adding one for a numbered paste if you take this.

